PR #12, opened 2017-11-21
Update travis testing for OpenMPI and MPICH - and move mpi4py v3.0.0

shuds13:
There is now a build matrix for different MPI libraries and pip install is used for mpi4py and PETSc, so these are separated from MPI libraries and should work correctly. Currently mpich and openmpi are tested. pip install mpi should now install v3.0.0, and so the develop version is no longer required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,29 @@ language: python
 python:
   - 2.7
   - 3.5
+  
+os: linux
+dist: trusty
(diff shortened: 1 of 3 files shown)


shuds13:
Updated with addition of conda/find_mpi.py diagnostic script that is referenced by travis - now travis passes.

PR #36, opened 2018-08-06
Inverse bayes example

jmlarson1:


shuds13:
You may want to change name of inverse_bayes_example.py to test_ inverse_bayes_example.py so it gets tested. Unless its deliberate?

jmlarson1:
I was just seeing how the merge went since the branch was made before the
recent API changes.

Perhaps I should have tried a merge master into it.

PR #58, opened 2018-08-17
Controller edits

dbindel:
Quickly discussed most of these with SH already.


jmlarson1:
@shuds13 can merge when ready.

PR #59, opened 2018-08-18
Feature/test lib e exceptions

shuds13:
Added unit tests for manager/working exceptions by mocking calls to manager_main/worker_main and forcing exceptions. Have not been able to test mpi abort options as I was not able to mock that


jmlarson1:
`FileNotFoundError` does not exist in Python2.7

PR #66, opened 2018-08-22
Feature/removing paused and cleaning api

jmlarson1:


jmlarson1:
@shuds13 Our coverage decreased because we lost the test on worker exceptions when the manager is automatically rank 0.

PR #83, opened 2018-09-08
Feature/refactor controller

dbindel:
Work on refactoring controller module -- now have controller (containing JobController and MPIJobController) and balsam_controller.  There is also a separate MPIResources object inheriting from resources where the partitioning logic now lives (was previously part of the controller).  Moved kill and poll from controller into job (though there is still a kill in the controller that calls the job kill with a default timeout).

Simplified the kill mode logic: now everything goes through cancel, with a timeout argument; three cases:

timeout=None:  Send SIGTERM and then wait (possibly forever) for completion
timeout=0:  Send SIGKILL immediately and then wait for completion
timeout>0:  Send SIGTERM and wait for timeout; if timeout before completion, send SIGKILL and wait

Also older work adding three new modules:
- loc_stack: keeps a stack of locations
- timer: manages a stopwatch
- launcher: encapsulates subprocess stuff and wait-and-kill logic, can be tested separately from JobController

These were used to separate out logic that was (mostly) previously in the worker and the controller.


dbindel:
Additional interface changes in last batch of commits:

- New mpi_controller module where MPIController lives
- Merged Register into JobController

PR #99, opened 2018-11-14
Feature/improved aposmm

jmlarson1:
Steve,

What are your thoughts on taking this into master too?


shuds13:
Are you hoping for a 0.4.1 release or just putting in master - before pulling in the comms stuff?

PR #145, opened 2019-04-12
Hotfix/giving back batches to persistent workers

jmlarson1:
The example allocation function was not correctly giving batches of work back to persistent generators. 

By adding the default behavior of recording the time that a batch gets entered into the history (in the `gen_time` field), we can easily identify (and give back to the waiting persistent gen) all points in the last batch that was given.


jmlarson1:
I adjusted the regression test to check a little more about the process. The correct thing to check is that the true function values came back to the persistent gen. Should we do the function value inside of the gen func?

PR #163, opened 2019-05-15
Testing/updates

shuds13:
Adjusted header docstrings for regression tests. Used a script so can easily be redone. 

Note description in chwirut tests are the same - is this correct?
Also check description in branin example.

So test_calc_exception.py is failing - it did not get run before as did not have the TESTSUITE_COMMS lines.

### DIFF
--- a/libensemble/libE_manager.py
+++ b/libensemble/libE_manager.py
@@ -281,6 +281,7 @@ class Manager:
 
         if isinstance(D_recv, WorkerErrMsg):
             self.W[w-1]['active'] = 0
+            self._kill_workers()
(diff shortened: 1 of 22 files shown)


jmlarson1:
I added correct documentation to the describe the regression tests using chwirut.

The test_calc_exception is hanging with more than 1 workers. It appears that the manager receives its first exception and exits without posting a nonblocking receive and an exit signal to the workers.

jmlarson1:
I removed tcp/local from chwirut tests. There was a issue with an error arising from the fact that is_master is true for all workers with tcp & local use cases.

PR #181, opened 2019-06-19
Testing/run tests mac os compatibility

jlnav:
Resolves an issue with run-tests.sh that prevented the regression tests from being performed on macOS.


jmlarson1:
I let Steve make the call on this. I think whichever one of `sed` or `grep` is more universal should be used.

shuds13:
To clarify, I'm fine using sed which was the main change. The issue here is that MAC does not support date +%s.%N

PR #207, opened 2019-07-18
xSDK - Fix/print to logger

jlnav:
Converts several hardwired stdout and stderr print statements into logger statements for closer adherence with xSDK policy M11. These libE and Manager outputs can now be turned off programmatically or read within ensemble.log.

Please let me know if I approached this incorrectly or if this change isn't especially necessary.


shuds13:
I think the raising of exceptions may be a special case that we want to go to at least standard error. Any other opinions on this.

Possibly we could use the logger and have a streamhandler that sends messages of type Error to standard error as well as log.

PR #210, opened 2019-07-19
Experimental/sim dir reserve

### LINKED ISSUE #146: Every worker copies/enters sim_dir even when they never do sim work
Related: What if there are gen_dirs and a worker moves between doing sim/gen work?

jlnav:
Closes #146 

Workers now only initialize and use a sim_dir if they perform sim work.


jlnav:
Good catch. I was using the test_branin_aposmm_nlopt_and_then_scipy.py calling script with the sim_funcs/branin sim_dir, which produced branin_sim-worker_1/ directories.

In the future, if we implement gen_dirs, we'll need to be more specific with directory naming. In the meantime, your suggestion is good for avoiding possible redundancy.

shuds13:
This looks good to merge now when tests complete.

PR #214, opened 2019-07-29
Feature/serial check inputs

### LINKED ISSUE #65: Allow for a serial check_inputs for users

jlnav:
Closes #65 

After including `from libensemble.libE import check_inputs` in their calling script, users can check any combination of libEnsemble settings dictionaries.

For example: `check_inputs(sim_specs=simspecs, gen_specs=genspecs, exit_criteria=exitc)`

Please let me know if I should include anything else.


libensemble/libE.py
```diff
@@ -498,3 +503,28 @@ def check_inputs(libE_specs, alloc_specs, sim_specs, gen_specs, exit_criteria, H
         # Check dimensional compatibility of fields
```
jmlarson1: I don't see why the prior history cannot contain unreturned points. If a point was previously generated, libE shouldn't have any trouble handling that. We need to mark those point that are 'given' as not 'given'. We just need to be careful that the time given, gen_worker, and other things are logical when restarted.

shuds13: check_inputs could give a warning if H0 does not meet the criteria, but if we want to actively fix up H0, where should we do that? We could do that in History.__init__, or do you want a separate utility function? I see the check_inputs routine as just failing, and possibly warning, if something is wrong.

shuds13: I'm inclined towards, erroring if H0 does not meet the criteria in check_inputs for now, and then making a new issue for H0 fixup. Where criteria is no values that are given, but not returned. Then we can pull this in.

PR #249, opened 2019-09-13
Debugging/balsam no file Resolution

jlnav:
(hopefully) Resolves intermittent Balsam File Not Found error.

Each Travis job has a generated hostname similar to "travis-job-210bb586-bfc7-44fa-abd5-8d46e097ebfa". If this hostname contains 'cc', Balsam will think it's running on Cooley!

This implements another ugly hack to Balsam to not be so strict. Unfortunately, the method to change Travis job hostnames via .travis.yml appears to break PostgreSQL on installation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,6 @@ matrix:
 services:
     - postgresql
 
-addons:
-  hostname: libensemble-travis
 
 # matrix:
 #  allow_failures:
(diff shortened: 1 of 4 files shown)


jmlarson1:
I was getting a lot of these... more than would seem likely for the suspected cause. 

Sadly, and I'm not smart enough to calculate the probability of getting consecutive `c`'s in a hash of that form. :-)

PR #251, opened 2019-09-26
Debugging/travis clang

jlnav:
Currently, there appears to be an issue with the Xcode SDK on every macOS Mojave (10.14) build on Travis that causes Balsam's installation to fail. For now, this sets Travis to test on macOS 10.13.

This also updates the downloaded miniconda version and slims down the macOS configuration process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,13 +65,12 @@ install:
         mkdir ../sdk; tar xf MacOSX10.14.sdk.tar.xz -C ../sdk;
         COMPILERS=clang_osx-64;
         MUMPS=mumps-mpi=5.1.2=haf446c3_1007;
-        echo "export CONDA_BUILD_SYSROOT=/Users/travis/build/Libensemble/sdk/MacOSX10.14.sdk" > setenv.sh;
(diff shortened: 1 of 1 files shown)


jmlarson1:
Any idea why there is a 1.5% drop in code coverage? I'm confused by the coveralls link that shows 1 of 4 (why more than 1?) versions of `balsam_controller.py` have zero coverage.

jlnav:
It appears in Coveralls' Tree listing that there is a copy of each file for each job in a build on Travis. The version that has zero coverage corresponds to the Python 3.5 job, where we don't test Balsam.

Comparing [this build](https://coveralls.io/builds/25936217) and a [previous develop build](https://coveralls.io/builds/25741805), I'm a bit puzzled about how overall and file-specific coverage is determined by Coveralls. Especially considering that the previous develop build has all four versions of `balsam_controller.py` covered for some reason. This new coverage _might_ be more accurate.

jmlarson1:
Yes, it is curious. This shouldn't hold up merging (if everything else is okay)

Issue #233, opened 2019-08-27
Build and run at scale on Cori


shuds13:
Port libEnsemble to Cori and run forces tests at scale.
  
Cori should also be used as an HPC test platform.

shuds13:
libEnsemble has been run with forces on 128 nodes on Cori and in several configurations. Auto-resources is working.

PR #257, opened 2019-10-03
Feature/port cori

shuds13:
libEnsemble has been ported to Cori in support of #233 

Runs have been performed at scale with the forces test in several configurations and with auto-resources.


libensemble/resources.py
```diff
@@ -150,13 +150,23 @@ def get_MPI_variant():
...
-            return 'mpich'
-        return 'openmpi'
+        try:
+            # Explore mpi4py.MPI.get_vendor() and mpi4py.MPI.Get_library_version() for mpi4py
+            try_mpich = subprocess.Popen(['mpirun', '-npernode'], stdout=subprocess.PIPE,
+                                         stderr=subprocess.STDOUT)
+            stdout, _ = try_mpich.communicate()
+            if 'unrecognized argument npernode' in stdout.decode():
+                return 'mpich'
+            return 'openmpi'
```
jmlarson1: Steve,
Are we assuming only mpich or openmpi? (Seems safe. Should we make it explicit somewhere?)

shuds13: The process looks for other launch commands like aprun/jsrun. If it comes down to mpirun, this is testing whether its using mpich or openMPI as a base, which have different options to mpirun. So yes, currently mpirun is assumed to be one of those. Intel MPI, for example, is using the mpich commands. I'm not sure what other implementations there may be.

Issue #260, opened 2019-10-07
gen_specs['in'] can't be empty for non-persistent gen_f's


jlnav:
Can be reproduced with the tutorial's `tutorial_calling.py` within `examples/tutorials` by removing the `'sim_id'` field from `gen_specs['in']`

Raises: `AssertionError: Allocation function requested rows=range(0, 20) be sent to worker=1, but requested no fields to be sent.` error in `_check_work_order()` in `libE_manager.py` on line 208.

With the associated assertion removed, raises: `IndexError: index 0 is out of bounds for axis 0 with size 0` in `_send_work_order()` in `libE_manager.py` on line 222.

Changing `range(len(H))` on line 81 from the `give_sim_work_first.py` alloc_f to `[]` on line 81 causes the tutorial code to run as expected but results in the `test_branin_aposmm_nlopt_and_then_scipy.py` test hanging.

jmlarson1:
I believe the best resolution is to have the default allocation function (`give_sim_work_first`) only give `gen_specs['in']` if there is something to send. Resolved by 
https://github.com/Libensemble/libensemble/pull/261

jlnav:
Closed by #261

PR #261, opened 2019-10-08
Changing default alloc func to only give gen_specs['in'] if it's a fi…

jmlarson1:
…eld with length


jmlarson1:
Closes https://github.com/Libensemble/libensemble/issues/260

Issue #269, opened 2019-10-29
[Tentative] Add tests/assertions for allowed x_specs fields, implement x_specs['user']


jlnav:
Related to #266 

'x' refers to either 'sim', 'gen', or 'alloc'.

Considering what data is necessarily passed to user functions, how to best differentiate between libE-reserved fields and user-defined-fields, and how to support sim_f/gen_f's with differing output data-types, we seem to now be leaning towards option 2 from #266 .

This will require several updates:

- [ ]  Implement checks and assertions that libE/user fields are defined in their proper structures. For example, keep user-defined sim fields in sim_specs['user'] 

- [ ]  Construct tests for these assertions, also update formatting for other tests.

- [ ] Update documentation, including examples and tutorials

Feel free to update this issue and/or offer suggestions for proceeding.

jmlarson1:
@jlnav I think this is somewhat addressed by PR #271, except the update of documentation. I'm not sure what other assertions we might need, other than checking that `x_spec` keys match a pre-defined list. (After compiling that list in libE.py, I think we should consider moving some things to `libE_specs`, but that is later.)

jlnav:
Presumably closed by #271

PR #271, opened 2019-10-31
Feature/structured specs

jmlarson1:
I have implemented the usage of 'user' for the unit/regression tests and example sim/gen funcs.

I plan on moving some of the contents of `sim_specs` and `gen_specs` to `libE_specs`, but I will do that in a different PR


jmlarson1:
This PR is in response to Issue #269

PR #278, opened 2019-11-06
Docs/Platforms - 1st Revision

jlnav:
This is the first round of revisions for the getting-started docs, featuring an index for background information on HPC systems, a page for the example job-submission scripts, some new diagrams to depict libE-process node affinity, and various other changes.

Comments and feedback are appreciated as always.

edit: contains changes for tutorial reflecting 'user' requirement.


jlnav:
> Is it possible to include the source files that make the .png in /docs/images/?

Sure, but I made the diagrams with [draw.io](https://draw.io/), source-files are saved as `.drawio` files, not exactly an open format.

PR #282, opened 2019-11-08
Docs/reorgainzing pdf

jmlarson1:


shuds13:
The pdf mostly looks good and I think the overall structure works. There are some small points. I'm also wondering about page breaks and the Bibliography?

I'm not sure the structure works so well for the html. For example, I like to have more quick-links (including the quick-start), a link to the libE API (which is kind of hidden in this format), and I don't really think the "For Infomration" and "In Depth" are really descriptive of what they contain. A discussion is probably best for this.

PR #283, opened 2019-11-11
Testing/balsam coverage adjust

jlnav:
Small Balsam coverage adjustment. Removes Balsam-source edit in favor of adding coverage configuration in the app-submission process.


jmlarson1:
Failing on some platforms.

jlnav:
Balsam app file-paths are split assuming the first element is the python file to execute with all following elements being arguments:

```
args = path.split()
app = args[0]
if app.endswith('.py'):
...
```

 This complicates integration with our coverage tracking method, which invokes the coverage module and specifies a config file before the to-execute python file is listed:

`path = ' -m coverage run --parallel-mode --rcfile=./libensemble/tests/regression_tests/.bal_coveragerc /home/travis/libensemble/libensemble/tests/regression_tests/script_test_balsam_hworld.py'`

Issue #235, opened 2019-08-27
Assess nprocesses as libE_specs field name


shuds13:
'nprocesses' [int]:
        Number of worker processes to spawn (in local/tcp modes)

I propose this would be better named nworkers, as it would seem less ambiguous, unless there is a counter-argument.

jlnav:
This change makes sense, and would improve lines like:

`libE_specs = {'nprocesses': nworkers, 'comms': 'local'}` in `common.py` and `nworkers = libE_specs['nprocesses']` in `libE.py`.

Also, when using `parse_args()` manually, `nprocesses` is specified as `--nworkers #` anyway.

PR #287, opened 2019-11-14
Move some utilities from regression_tests/common.py to libensemble/utils.py

jlnav:
I think I covered all the tests and examples that use these utilities. We'll know right away if anything breaks.


jmlarson1:
Any thoughts on why `ssh_parse_args` is no longer covered? 
https://coveralls.io/builds/27002755/source?filename=libensemble%2Futils.py#L297

jlnav:
> Any thoughts on why `ssh_parse_args` is no longer covered?
> https://coveralls.io/builds/27002755/source?filename=libensemble%2Futils.py#L297

It wasn't covered in the first place. The `parse_args` routines were in files in `/tests`, all of which are excluded from coverage.

Now we know we need to either develop tests or drop support for our ssh comms capabilities.

jlnav:
Added commit to address #235 . Can be reverted if not worth implementing.

PR #289, opened 2019-11-18
Renaming per_worker_streams

jmlarson1:
Renaming per_worker_streams, add a doc string, and changing an argument name.

### DIFF
--- a/libensemble/utils.py
+++ b/libensemble/utils.py
@@ -364,7 +364,7 @@ def save_libE_output(H, persis_info, calling_file, nworkers):
 
 def add_unique_random_streams(persis_info, size):
     # Used to create a random number stream for the libE manager and workers
-    # (when size = 1 + num_workers). Stream i is initialized with seed=i. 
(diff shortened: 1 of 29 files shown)


libensemble/utils.py
```diff
@@ -362,8 +362,10 @@ def save_libE_output(H, persis_info, calling_file, nworkers):
 # ===================== per-worker numpy random-streams ========================
 
 
-def per_worker_stream(persis_info, nworkers):
-    for i in range(nworkers):
+def add_unique_random_streams(persis_info, size):
+    # Used to create a random number stream for the libE manager and workers
+    # (when size = 1 + num_workers). Stream i is initialized with seed=i.
```
jlnav: How about "Creates ``size`` random number streams for the libE manager and workers, where ``size`` is ``num_workers`` + 1. Stream i is initialized with seed i." ?

shuds13: I'm  happy to do the wording change, it helps highlight multiple streams

PR #290, opened 2019-11-18
move report_manager_exception back to libE.py

jlnav:


jmlarson1:
@shuds13 is b143637 what you were asking about?

PR #291, opened 2019-11-18
Feature/fixes for rel 053

shuds13:
Updated forces example and small fixes. This should leave only documentation for release.


shuds13:
Ok, I'm doing the flake - dollar in the forgot to flake box.

PR #297, opened 2019-11-20
Docs/example calling scripts

jlnav:
Adds example calling scripts, also updates tutorial and adds literal-include captions throughout.


jlnav:
> The layout and captions look good. A few thoughts.
> 
> Job-controller - not sure this is the best example as its a bit of an artificial test.
> And all the stuff at the bottom is for results testing. I think forces might be a better
> example to showcase the job-controller. I've also restructured it to use the parse_args now.
> I would like to keep the MPIController/BalsamController option as it shows, hopefully, how
> easy to switch between them.
> 
> APOSMM: Remove use of support
> 
> You may also want to note that the first two are just random sampling, while the third
> is performing optimization.

Suggested changes made

PR #298, opened 2019-11-20
Docs/shudson fixes

### LINKED ISSUE #232: Add installation section to docs
In addition to whats in the getting started guide, this will provide instructions for different systems - eg. Theta, Summit. It will include spack instructions and should also mention if modules are available on those systems.

Some considerations will be:
Building with Conda, and using intel packages where available.
Installing mpi4py in right way to pick up correct system MPI

Spack install docs will include build instructions for different platforms, links to various yaml files (packages, modules, config) to support and possibly back-end build scripts where relevant.

shuds13:
Documents calc_status and adds Cori guide. I will keep this branch open but I want to bring these changes in.  Should resolve #232

### DIFF
--- a/docs/data_structures/calc_status.rst
+++ b/docs/data_structures/calc_status.rst
@@ -51,7 +51,7 @@ Available values:
     :start-after: first_calc_status_rst_tag
     :end-before: last_calc_status_rst_tag
 
-The corresponding messages printed to the libE_stats.txt file are:
+The corresponding messages printed to the ``libE_stats.txt`` file are:
 
 ..  literalinclude:: ../../libensemble/message_numbers.py
     :start-at: calc_status_strings
--- a/docs/data_structures/calc_status.rst
+++ b/docs/data_structures/calc_status.rst
@@ -1,0 +1,74 @@
+.. _datastruct-calc-status:
+
+calc_status
+===========
+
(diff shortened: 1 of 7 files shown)


docs/data_structures/calc_status.rst
```diff
@@ -0,0 +1,74 @@
+.. _datastruct-calc-status:
+
+calc_status
+===========
+
+The ``calc_status`` is an integer attribute with a corresponding description that can be used in :ref:`sim_f<api_sim_f>` or :ref:`gen_f<api_gen_f>` functions to capture the status of a calcuation. This is returned to the manager and printed to the ``libE_stats.txt`` file. This is not currently used by the manager, but can still provide a useful summary in libE_stats.txt. The user determines the status of the calculation, as it could include multiple application runs. It can be added as a third return variable in sim_f or gen_f functions.
```
jmlarson1: We say "This is not currently used by the manager" but isn't it used when a worker stops being persistent (on it's own)?

shuds13: Ok, in that case I will also move these into the calc_status section in message_numbers.py and give a description (which will go to libE_stats.txt). I will change this.

Looks like its only used in uniform_or_localopt.py currently.

shuds13: I've changed this now.

docs/platforms/cori.rst
```diff
@@ -0,0 +1,125 @@
...
+Debugging Strategies
+--------------------
+
+View the status of your submitted jobs with ``squeue`` and cancel jobs with
+``scancel [Job ID]``.
+
+Additional Information
+----------------------
+
+See the LCRC Bebop docs here_ for more information about Cori.
```
jlnav: "LCRC Bebop" -> "NERSC Cori"

shuds13: Nice spot!

PR #303, opened 2019-11-25
Docs/update example sub scripts

shuds13:


examples/job_submission_scripts/summit_submit_mproc.sh
```diff
@@ -0,0 +1,43 @@
...
+export EXE=run_libe_forces.py
+
+# Communication Method
+export COMMS="--comms local"
+
+# Number of workers.
+export NWORKERS="--nworkers 4"
+
+# Wallclock for libE.  (allow clean shutdown)
+#export LIBE_WALLCLOCK=25 # Optional if pass to script
```
jmlarson1: Do we want to include this?

shuds13: We probably should. But I will do this in my next pull request as it will be supplied as a positional argument that is not recognised by parse_args. In this version that will be an error. Following my next pull request parse_args will ignore it (branch feature/fix_utils) so the user script can pick it up. It would be nice for parse_args to process a --wall option, but we will have to involve exit_critera as another output. Unless we make wall_clock part of libE_specs.

Issue #267, opened 2019-10-18
Consider restructuring sim and gen directory creation


shuds13:
The following applies both to sim and gen.

Currently, a directory sim_worker_<worker_id> (or gen_worker_<worker_id>) is created for each each worker, each copying the contents of a specified sim (or gen) input directory. For each point (or each sim call) evaluated by the worker, the contents of this directory is either overwritten, or the user is responsible, via their sim func, to create sub-directories in which to run each job - involving a second copy of input data (or possibly a sim link). This does allow the user to name the directory as they wish.

Also, the separate worker directories can potentially live on distributed storage resources (e.g. on-node SDD). And input data copied to the worker directory can be re-used between sim calls.

Note, that in some cases there is a sim and gen worker directory for the same worker.

This two step process may place an unnecessary burden on the user to set up their own simulation sub-directories in the sim func, and in cases where local storage resources are not used, the separate worker directories may not be needed.

A more intuitive approach may be to focus on copying/referencing the input data for each point in H, or each sim (or gen) call, rather than for each worker. Many users may not require to name each directory, and a random hash could be used. libensemble can then place the user and run the sim func  inside that directory. Also, the worker directories could be optional, and would just be worker1/worker2..

A downside of this approach may be reduced flexibility. If the user does want to name their own directories inside the sim func. Or if they want to store shared input in the worker level directory. Provisions could be made for these options while having a simpler default. The job_controller (as currently named) could also help with a jobctl.create_workdir(name) function once a directory name is created in the user sim func. This function could also potentially take other options for data copying/referencing.

jlnav:
Moving some comments and progress to Issue from Pull Request.

TODO:

- [ ] Per-calculation directories differentiated with sim IDs instead. Or, spec to allow user to choose.

- [ ] Per-calculation directories created by default, but can be disabled

- [ ] Rename specs options: 'sim_dir' -> 'input_dir', 'sim_dir_prefix' -> 'ensemble_dir_path', etc.

- [ ] Setting for user to list which files to symlink, which to copy.

- [ ] Worker-directories don't need input_dir prefix, only worker-number.

jlnav:
Tentatively closed by #334

PR #304, opened 2019-11-25
Create sim_dir for each sim call instead of for each worker by default.

jlnav:
Addresses #267

New approach for creation of simulation directories. Now, _per-calculation worker directories are created by default_, with a new ``libE_specs`` option "``do_worker_dir``" for specifying per-worker directories.

This also includes changes to forces and a regression test, and a new regression test.

Please provide feedback and suggestions for developing or adjusting this feature. I recommend checking out the branch and trying it out. Thank you!


jlnav:
New libE_specs parameter, `do_worker_dir` created. Enables "traditional" worker directories.

jlnav:
Closing for now

shuds13:
This is part way there, but still quite a few scenarios we need to consider.

My proposed default is to have a copy of the sim dir (or gen dir), or a referenced copy (see below), for each calc, whether or not worker directories are used. If the per_worker option is specified, then it would put these under worker directories. This may be debatable though, as I said in the initial proposal, users may wish to name their directories in the sim func. A libE provided helper function could be used to create the named sim_directory and place them in it, for instance. Alternatively, they could provide to libE a formula to construct directory name from sim parameters, and then libE can create and still place you in the sim directory. Either way, they may wish to do this irrespective of whether they are using worker directories.

I also think it is best to create an upper level directory for the user's ensemble. Definitely if not using worker dirs, but consider when they run 1000 worker ensemble, they will want this anyway.

So my view would be (provisionally calling the upper level directory 'ensemble'. Directory names being either appended by a hash, or H row number/range.

Default dir structure.:

```
-ensemble/
    -gen_0/ (when that needs a directory)
    -sim_0/
    -sim_1/
    -sim_2/
    -sim_3/
  ...
```

If worker_dir option is provided in libE_specs.....

```
-ensemble/
    -worker1/
        -gen_0/ (when that needs a directory)
        -sim_0/
        -sim_2/
    -worker2/
        -sim_1/
        -sim_3/
```

By a referenced copy, I mean the user might want to use symoblic links in some cases to avoid making copies of large files, to either all, or possibly some, or the files in simdir. So in that case, a per worker scheme, presuming the workers might use isolated storage locations, would make a copy of sim dir into the worker directories. And then each sim directory (within that) can use symoblic links. I'm not totally sure on the best way to specify the sym links v hard copy. But I'm thinking the best default is that the user sim func is run with current working directory inside the created sim directory.

This still leaves the question of whether some people want to stick with having worker directories without separate sim dirs, if they are happy to overwrite calculations, and at least workers will be doing only one at a time. Or if they want more control over creating their directory structures. We may want to discuss this and the best set of options to express it.

Please debate any of this, but if the above scheme is implemented the first change is simply creating a top level dir and creating sim dirs within the worker dirs as a default.

So options (provisionally named) may be:

```
worker_dirs?            default False
sym_link_to_input?      default False? (If worker_dirs is true - will copy to each worker dir).
sim_dirs?               default True (if false user resp. for creating their own - as now! We could provide helper func.)
sim_dir_numbering?      Options row/hash/possibly user scheme (based on sim input).
```

Note that use of sim dirs might want to be generalised to calc dirs  (in cases where a gen dir is necessary). I doubt we need separate options - they will only be created if they are used.

jlnav:
Suggested directory structure can now be recreated with a combination of libE_specs options ``sim_dir``, ``do_worker_dir``, and ``sim_dir_prefix``. See new ``test_worker_sim_dirs.py`` regression test, and adjust ``do_worker_dir`` 's value to produce either dir structure.

jlnav:
Currently developing symlinking capability.

shuds13:
So the structure looks right.

We might want to discuss the options/terminology. 

I think the 'sim_dir' in this case might want to prefix at the sim dir level and the workers dirs probably just worker1, worker2. Any thoughts?

I'm inclined to having ensemble/ level dir always - or atleast as default. If we have the option of naming, the key should be something other than "sim_dir_prefix", remembering that
it could contain gen dirs also. Maybe "ensemble_name". This could provide a trivial way for users' to re-run in the same directory without deleting previous ensembles. Possibly, we might also think about appending a count automatically if the directory exists when they run.

shuds13:
I think this is working right when worker directories are not used. But when worker directories are used the sim link is at the worker directory level. I think we need to copy the data into the worker directories and then sim link to that copy from within the sim directories. This covers the case when worker directories are in different localised filesystems.

jlnav:
Preliminary Symlinking capability developed. Source files are copied over for each worker-directory, with each nested calc-directory containing symlinks to these copies. Symlinking can be enabled/disabled. If not using worker-directories (for whatever reason), sim-dir copies symlink to the original source. Perhaps they should symlink to a single copy under the ensemble directory instead.

jlnav:
The status of new features and paradigms are as follows:

Suppose we have a sim (or "input"; I'm considering renaming) directory `./test_dir`:

```
- test_dir/
   - copythis.txt
```

_Only_ setting `libE_specs['sim_dir'] = './test_dir'` produces a directory structure like:

```
- ensemble/
   - copythis.txt
   - test_dir_worker1-51e48b3e/
      - copythis.txt
      - output
   - test_dir_worker2-d59954de/
   - test_dir_worker3-5c1be408
   ...
```

Note that an input-file copy was placed in the top-level ensemble directory _and_ inside each calculation directory.

The top-level directory is named "ensemble" by default, but can be adjusted with:

` libE_specs['sim_dir_prefix'] = './my_ensemble'`

This parameter might need renaming too.

Suppose we define the following settings:

```
libE_specs['sim_dir'] = './test_dir'
libE_specs['do_worker_dir'] = True
libE_specs['sim_dir_prefix'] = './my_ensemble'
```

This will produce a directory structure like:

```
- my_ensemble/
   - test_dir_worker1/
      - copythis.txt
      - sim_1e5ac826/
         - copythis.txt
         - output
      - sim_5e36d369/
      - sim_c495a4d6/
      ...
   - test_dir_worker2/
   - test_dir_worker3/
   ...
```

The input-files are copied for each worker and each calculation similarly to above.

Suppose we now define an additional new setting and set `do_worker_dir` to `False`:

` libE_specs['sym_link_to_input'] = True `

This will produce a _per-calculation_ directory structure like:

```
- my_ensemble/
   - copythis.txt
   - test_dir_worker1-51e48b3e/
      - *copythis.txt
      - output
   - test_dir_worker2-d59954de/
   - test_dir_worker3-5c1be408
   ...
```

Where the input-files is copied once for the ensemble, and each per-calculation symlink refer to this copy.

If we define all four of these settings together:

```
libE_specs['sim_dir'] = './test_dir'
libE_specs['do_worker_dir'] = True
libE_specs['sim_dir_prefix'] = './my_ensemble'
libE_specs['sym_link_to_input'] = True
```
This will produce a directory structure similar to:

```
- my_ensemble/
   - test_dir_worker1/
      - copythis.txt
      - sim_1e5ac826/
         - *copythis.txt
         - output
      - sim_5e36d369/
      - sim_c495a4d6/
      ...
   - test_dir_worker2/
   - test_dir_worker3/
   ...
```

Where the input-files are copied for each worker, and each nested calculation-directory contains symlinks to these per-worker copies.

This implementation was developed with that idea that each ensemble-directory or each worker-directory is encompassing, with no outward symlinks or data-dependencies in case these directories are output to separate filesystems. 

Additional features that need developing (but probably in another branch):
   - An option for direct-to-source symlinks, or default behavior can be changed. Furthermore, 
   - A disabling setting for per-calculation directories
   - Customize the differentiation mechanism for per-calculation directories

I recommend checking out the `test_worker_sim_dirs.py` regression test and experimenting with different combinations of settings. Comment-out the `shutil.rmtree()` routines within the `is_master` blocks to observe the ensemble directories.

Please let me know of any questions or concerns. I'm considering implementing additional features in a new branch, however.

Thank you!

jlnav:
ALSO, another new libE_specs feature:

`libE_specs['input_files'] = ['copythis.txt', 'copythat.py']`

Specify exactly which files out of the input-directory to copy/symlink over, or copy/symlink all by default if not specified.

jlnav:
> ## Pull Request Test Coverage Report for [Build 2505](https://coveralls.io/builds/27690784)
> 
>     * **56** of **56**   **(100.0%)**  changed or added relevant lines in **2** files are covered.
> 
>     * **22** unchanged lines in **2** files lost coverage.
> 
>     * Overall coverage increased (+**0.08%**) to **94.77%**
> 
> 
> Files with Coverage Reduction 	New Missed Lines 	%
> [libensemble/gen_funcs/persistent_aposmm.py](https://coveralls.io/builds/27690784/source?filename=libensemble%2Fgen_funcs%2Fpersistent_aposmm.py#L241) 	10 	95.27%
> [libensemble/utils.py](https://coveralls.io/builds/27690784/source?filename=libensemble%2Futils.py#L303) 	12 	90.32%
> Totals 	[Coverage Status](https://coveralls.io/builds/27690784)
> Change from base [Build 2456](https://coveralls.io/builds/27564521): 	0.08%
> Covered Lines: 	11363
> Relevant Lines: 	11875
> ##### 💛 - [Coveralls](https://coveralls.io)

Fancy!

jmlarson1:
I checked the "detailed" reports from coveralls. I was hoping that it would let me know if it was just a "comms" issue when our coverage drops.

jlnav:
TODO:

- [ ] Per-calculation directories differentiated with sim IDs instead. Or, spec to allow user to choose.

- [ ] Per-calculation directories created by default, but can be disabled

- [ ] Rename specs options: 'sim_dir' -> 'input_dir', 'sim_dir_prefix' -> 'ensemble_dir_path', etc.

- [ ] Setting for user to list which files to symlink, which to copy.

- [ ] Worker-directories don't need input_dir prefix, only worker-number.

Also, closing for now until more features added and polished.

PR #317, opened 2019-12-06
Regression Test for libE_worker cProfile

jlnav:
Should cover a decent chunk of lines in libE_worker involving the profile_worker feature.


jmlarson1:
Looks good. Any thoughts on getting to 100% in the libE_worker and libE_manager?

jlnav:
We're evidently making progress. I just realized that libE_manager also contains a profiling mechanism that needs covering.

PR #320, opened 2019-12-11
Feature/subsecond time

jlnav:
The Timer module now measures in milliseconds behind-the-scenes. ``libE_stats`` displays additional millisecond values for Start and End times:

``Start: 2019-12-11 10:27:50.227 End: 2019-12-11 10:27:50.733``

However, for compatibility with current timing mechanisms ``timer.elapsed``  and ``timer.total`` times are returned as seconds. If necessary, I can implement additional settings or parameters for milliseconds if we want to be more precise throughout the project.

### DIFF
--- a/libensemble/tests/unit_tests/test_timer.py
+++ b/libensemble/tests/unit_tests/test_timer.py
@@ -48,7 +48,7 @@ def test_timer():
 
     time_end = time.time() - time_start + time_mid
 
-    assert total1>= 1 and total1 <= time_end, \
+    assert total1 >= 1 and total1 <= time_end, \
         "Check cumulative timing (active)."
     assert timer.total >= 1 and timer.total <= time_end, \
         "Check cumulative timing (not active)."
--- a/libensemble/tests/unit_tests/test_timer.py
+++ b/libensemble/tests/unit_tests/test_timer.py
@@ -27,7 +27,8 @@ def test_timer():
     # Use external wall-clock time for upper limit to allow for system overhead
     # (e.g. virtualization, or sharing machine with other tasks)
     # assert (e1 >= 0.5) and (e1 <= 0.6), "Check timed sleep seems correct"
-    assert (e1 >= 0.5) and (e1 < time_mid), "Check timed sleep within boundaries"
+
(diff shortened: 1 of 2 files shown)


jmlarson1:
@jlnav  Is it possible to replace the two uses of `time` with `datetime`? For example, using `datetime.strftime` and `time_tuple` which should be like `time.localtime()`, see:
https://docs.python.org/3/library/datetime.html#datetime.date.timetuple

Thoughts?

jlnav:
> @jlnav Is it possible to replace the two uses of `time` with `datetime`? For example, using `datetime.strftime` and `time_tuple` which should be like `time.localtime()`, see:
> https://docs.python.org/3/library/datetime.html#datetime.date.timetuple
> 
> Thoughts?

Good point. Fixed.

jmlarson1:
I don't think this is working as expected. When I run, I get the following output: 

````
Worker     2: Calc     0: sim Time: 0.00 Start: 2019-12-11 00:00:00.405 End: 2019-12-11 00:00:00.406 Status: Not set
Worker     3: Calc     0: sim Time: 0.00 Start: 2019-12-11 00:00:00.406 End: 2019-12-11 00:00:00.406 Status: Not set
Worker     3: Calc     1: sim Time: 0.00 Start: 2019-12-11 00:00:00.409 End: 2019-12-11 00:00:00.409 Status: Not set
Worker     2: Calc     1: sim Time: 0.00 Start: 2019-12-11 00:00:00.408 End: 2019-12-11 00:00:00.408 Status: Not set
...
...
````
The time is not correct, and the `Status` is `Not Set` for some reason.

shuds13:
It will be Not Set unless you are using the calc_status feature. I may be inclined to only write the milliseconds to file if run-times are smaller than some cut-off, as these lines can get very long - esp. if you include job run-times (as well as calc times).

shuds13:
I'm also getting milliseconds only on my laptop (running forces test).

jlnav:
Sorry about that. I think I just fixed the milliseconds-only issue.

shuds13:
Yes, the time issue seems to be fixed now.

Issue #338, opened 2020-01-17
'save_every_k' should replace existing files as a default


jmlarson1:
There is no reason to keep the old ones, since a later history will already contain anything printed out in an earlier history.

jmlarson1:
Solved by PR #339

PR #339, opened 2020-01-17
Feature/improving saves

jmlarson1:
There is no reason to keep previous checkpointing files. 

Any later history will already contain anything saved in an earlier history.

This also prevents libE from filling up a disk when, for example, running with `libE_specs['save_every_k_sims'] = 1` and `sim_max` is large.


jmlarson1:
Addresses issue #338

Issue #285, opened 2019-11-13
Reorganization of files


jmlarson1:
We have resources files:
````
libensemble/env_resources.py
libensemble/mpi_resources.py
libensemble/node_resources.py
libensemble/resources.py
````
controller files
````
libensemble/balsam_controller.py
libensemble/controller.py
libensemble/mpi_controller.py
````
two util locations
````
libensemble/util/
libensemble/utils.py
````
as well as
````
libensemble/tests/regression_tests/common.py
libensemble/tests/regression_tests/support.py
libensemble/gen_funcs/support.py
libensemble/alloc_funcs/support.py
````

Do we have a coherent motivation for this structure?

shuds13:
I've been thinking about sub-dirs for resources and controllers. This will change the import API though (for controllers). Bear in mind we are thinking of renmaing that also, so we might want to make that change in one go!

Currently the util/ directory is, I think, internal modules rather than user API. We might want to consider that.

jlnav:
Further considerations from Slack discussion:
 - [ ]  Have two utils directories; one for libE-intended functions, one for external user-intended convenience functions
 - [ ] Reformat internal routines to start with underscores, autodoc

jlnav:
Closed by #344

PR #344, opened 2020-01-29
File Reorganization and Module Renaming

jlnav:
Addresses #285 and several other things we've discussed. I don't recommend pulling this in until more discussion about terminology and organization has occurred, but this should serve as a good proof-of-concept and method for exploring specific changes. Any release these changes are included in will probably justify a major release.

This is primarily a terminology and reorganization branch, and includes the following major changes:

1. The `/conda` directory is now `/install`, since most of it's contents aren't conda-specific.
2. All resources related python files are now in the `libensemble/resources` directory
3. All job-controller related python files are now in the `libensemble/executors` directory.
4. Incidentally, the Job Controller is now called the Executor, with `MPI_Executor()` and `Balsam_MPI_Executor()` variants. This was primarily done to differentiate what libEnsemble accomplishes within user-functions versus what other largely separate frameworks like Balsam accomplish.
5. The `launch()` function is now `submit()`, more effectively communicating that "launched jobs" aren't necessarily executed immediately or queued by libEnsemble itself.
6. **Jobs** are now referred to as **Tasks** throughout the docs and code; this differentiates them from job-submission routines and avoid the "sub-job" complication. With this understanding, on a HPC system libEnsemble is queued as a "job" which then submits "tasks" for immediate execution or submission through Balsam. "Tasks" may also more closely adhere to the understanding of "a unit of computational work".

This means that calling scripts and user functions now have implementations that resemble:

```
from libensemble.executors.mpi_executor import MPI_Executor
exctr = MPI_Executor()
exctr.register_calc(full_path=sim_app, calc_type='sim')
```
and
```
from libensemble.executors.executor import Executor
...
exctr = Executor.executor
task = exctr.submit(calc_type='sim', app_args=args)
task.poll()
...
```

So the major changes for existing code would include importing each executor from the `libensemble.executors` module, renaming each executor, and replacing `launch()` with `submit()`. Of course, no changes are needed for how object-instances within user-code are named.

This does not include any changes to `/utils` or `/libensemble/utils.py`

I understand that many of these changes are dramatic and could take some getting used to if approved. It would be perfectly acceptable to only approve a few (if any) included changes.

As usual, I would appreciate many comments, suggestions, and personal improvements. The docs and tests pass, but please let me know if I overlooked something.


docs/FAQ.rst
```diff
@@ -83,19 +83,20 @@ Also see https://software.intel.com/en-us/articles/python-mpi4py-on-intel-true-s
...
+reduce the number of processes running or to specify a fallback fabric through
+environment variables::
 
     unset I_MPI_FABRICS
     export I_MPI_FABRICS_LIST=tmi,tcp
     export I_MPI_FALLBACK=1
 
 Alternatively, libEnsemble can be run in central mode where all workers run on dedicated
-nodes, while launching all sub-jobs onto other nodes.
+nodes, while launching all sub-tasks onto other nodes.
```
jmlarson1: Why is this "sub-tasks"? Doesn't central mode mean launching "a unit of computational work"?

docs/known_issues.rst
```diff
@@ -4,17 +4,17 @@ Known Issues
 The following selection describes known bugs, errors, or other difficulties that
 may occur when using libEnsemble.
 
-* OpenMPI does not work with direct MPI job launches in mpi4py comms mode, since
-  it does not support nested MPI launches. Use either local mode or the Balsam job
-  controller.
+* OpenMPI does not work with direct MPI task launches in mpi4py comms mode, since
```
jmlarson1: Should this "launches" be "submissions"?

docs/logging.rst
```diff
@@ -4,8 +4,8 @@ Logging
 The libEnsemble logger uses the standard Python logging levels (DEBUG, INFO, WARNING, ERROR, CRITICAL)
 plus one additional custom level (MANAGER_WARNING) between WARNING and ERROR.
 
-The default level is INFO, which includes information about how jobs are launched
-and when jobs are killed. To gain additional diagnostics, set the logging level
+The default level is INFO, which includes information about how tasks are launched
```
jmlarson1: Is this "launched" correct?

docs/overview_usecases.rst
```diff
@@ -15,8 +15,8 @@ that can perform calculations. libEnsemble's work is governed by three routines:
 * :ref:`alloc_f<api_alloc_f>`: Decides whether ``sim_f`` or ``gen_f`` should be called (and with what input/resources) as workers become available
 
 Example ``gen_f``, ``sim_f``, ``alloc_f``, and calling scripts can be found in
-the ``examples/`` directory. In order to enable portability, a :doc:`job_controller<job_controller/overview>`
-interface is supplied for users to launch and monitor external scripts in their
+the ``examples/`` directory. In order to enable portability, an :doc:`executor<executor/overview>`
+interface is supplied for users to execute and monitor external scripts in their
```
jmlarson1: Should this "execute" be "submit"?

docs/platforms/cori.rst
```diff
@@ -110,10 +110,10 @@ With your nodes allocated, queue your job to start with four MPI ranks::
 
 This line launches libEnsemble with a manager and **three** workers to one
 allocated compute node, with three nodes available for the workers to launch
-user applications with the job-controller or a job-launch command.
+user applications with the executor or a job-launch command.
```
jmlarson1: "job-submission"?

docs/tutorials/executor_forces_tutorial.rst
```diff
@@ -295,18 +296,18 @@ Load output data from our job and return to the libEnsemble manager:
...
-libEnsemble features two variants of its job controller that perform identical
+libEnsemble features two variants of its executor that perform identical
 functions, but are designed for running on different systems. For most uses,
 the MPI variant will be satisfactory. However, some systems, such as ALCF's Theta
 do not support MPI launches from compute nodes. On these systems libEnsemble is
-run either on launch nodes or using a proxy launch mechanism to submit
-jobs from compute nodes. One such mechanism is a scheduling utility called
-Balsam_ which runs on a separate node. The Balsam Job Controller variant interacts
+run either on launch nodes or uses a proxy launch mechanism to submit
+sub-jobs from compute nodes. One such mechanism is a scheduling utility called
```
jmlarson1: Why "sub-jobs" here?

libensemble/tests/balsam_tests/readme.balsam_tests.txt
```diff
@@ -1,11 +1,11 @@
 List of balsam test jobs for libEnsemble.
 
 test_balsam_1__runjobs.py:
-Launches parallel parent job. Each task iterates over a loop launching sub-jobs
+Launches parallel parent job. Each job iterates over a loop launching sub-jobs
```
jmlarson1: Why jobs?

libensemble/tests/balsam_tests/readme.balsam_tests.txt
```diff
@@ -1,11 +1,11 @@
...
 and waiting for completion. Does 3 iterations by default.
 
 test_balsam_2__workerkill.py:
-As first test but each top-level task kills its middle job.
+As first test but each top-level jobs kills its middle job.
```
jmlarson1: "middle job"?

jmlarson1:
@jlnav  I agree with this change, but I see quite a few variants of "launch" and "job" that I believe need to be changed. Perhaps that's too large of a change to see in a single PR, so we're waiting for later? (If you have some nice way to take items 1-4 into a first PR and then I'd consider that. But don't spend time trying to split it out.)

For example, running 

````$ grep -ir --include=*.py --include=*.rst launch ````

from the base `libensemble/` directory still has 255 instances (down from 346 on the `develop` branch). Similarly, I get 320 hits on 

````$ grep -ir --include=*.rst job````

(down from 1337 on `develop`). 

Not all of those "jobs" might need to be changed, but I think a lot of them should be. 

Not sure if we want to change the `CHANGELOG` for consistency (I would think so).

File: docs/overview_usecases.rst (hunk 4 above)
Comment (jlnav) [reply docs/overview_usecases.rst]: For this context, I think I prefer "execute" since that is the eventual action of the executor and best describes for those reading the overview that libEnsemble can run their code.

Thoughts?

File: docs/logging.rst (hunk 3 above)
Comment (jlnav) [reply docs/logging.rst]: This should be "submitted".

File: docs/known_issues.rst (hunk 2 above)
Comment (jlnav) [reply docs/known_issues.rst]: Hm, I could see it as either "submissions" or "executions" depending if we want to emphasize the issues alongside the executor or just for OpenMPI. I might rewrite that issue a bit.

File: docs/FAQ.rst (hunk 1 above)
Comment (jlnav) [reply docs/FAQ.rst]: You're right, this should be "tasks".

File: docs/platforms/cori.rst (hunk 5 above)
Comment (jlnav) [reply docs/platforms/cori.rst]: Probably "task-submission", but I'll need to ask @shuds13 about if Cori can launch jobs from compute nodes to know how to best reword this section.

File: docs/tutorials/executor_forces_tutorial.rst (hunk 6 above)
Comment (jlnav) [reply docs/tutorials/executor_forces_tutorial.rst]: libEnsemble submits "tasks" to Balsam, which are launched as "jobs" since they're enqueued and (I believe) launched to separate allocations. This is definitely confusing and worth discussing in a meeting.

File: libensemble/tests/balsam_tests/readme.balsam_tests.txt (hunk 7 above)
Comment (jlnav) [reply libensemble/tests/balsam_tests/readme.balsam_tests.txt]: Balsam launches "jobs", and these tests don't use the executor, let-alone libEnsemble.

File: libensemble/tests/balsam_tests/readme.balsam_tests.txt (hunk 8 above)
Comment (jlnav) [reply libensemble/tests/balsam_tests/readme.balsam_tests.txt]: Similar reasoning as above.

**Comment by jlnav** [comment]
> @jlnav I agree with this change, but I see quite a few variants of "launch" and "job" that I believe need to be changed. Perhaps that's too large of a change to see in a single PR, so we're waiting for later? (If you have some nice way to take items 1-4 into a first PR and then I'd consider that. But don't spend time trying to split it out.)
> 
> For example, running
> 
> `$ grep -ir --include=*.py --include=*.rst launch `
> 
> from the base `libensemble/` directory still has 255 instances (down from 346 on the `develop` branch).
> 
> Similarly, I get 320 hits on
> 
> `$ grep -ir --include=*.py --include=*.rst job`
> 
> (down from 1337 on `develop`).
> 
> Not all of those "jobs" might need to be changed, but I think a lot of them should be.
> 
> Not sure if we want to change the `CHANGELOG` for consistency (I would think so).

Alright. if possible, perhaps file-organization can be split from renaming. I'll need to investigate this. I also suggest that these changes aren't made until some period after the ECP meeting so that we don't confuse ourselves and others by explaining modules that have changed names on develop, but not the official release or docs.

For "launch" instances, I think most now refer to HPC node-types, launching workers, Balsam uses of the word (including commands), comments, tests, and the launcher; a module of the job controller (executor).

For "job" instances, most now refer to Balsam and HPC uses. I'm finding more examples that should be replaced and doing so.

In regards to the changelog, I thought it would be best to keep it as-is, since the next release might have an entry like "the Job Controller is now called the Executor", for example.

**Comment by jlnav** [comment]
launch_time -> submit_time

**Comment by jlnav** [comment]
definition of terms page

**Comment by jlnav** [comment]
@shuds13 

These are all good points. Maybe we should only be defining terms specific to libEnsemble, it's not our responsibility to define words we don't use regularly or aren't confusing. Also, we're renaming terms to prevent overlap.

We can't change anything about Balsam using "job". Yeah, the one thing that needs clarification is "libE tasks become Balsam jobs".

**Comment by jlnav** [comment]
plotting, jobs -> tasks

File: docs/overview_usecases.rst (hunk 4 above)
Comment (shuds13) [reply docs/overview_usecases.rst]: I think 'scripts' should be 'applications'.

File: docs/platforms/cori.rst (hunk 5 above)
Comment (shuds13) [reply docs/platforms/cori.rst]: I would say:
user applications (via the executor or a direct run command such as mpiexec)

File: docs/tutorials/executor_forces_tutorial.rst (hunk 6 above)
Comment (shuds13) [reply docs/tutorials/executor_forces_tutorial.rst]: I think it should be tasks.

File: libensemble/tests/balsam_tests/readme.balsam_tests.txt (hunk 7 above)
Comment (shuds13) [reply libensemble/tests/balsam_tests/readme.balsam_tests.txt]: Yep, I think we can leave this alone as its just Balsam tests.

File: libensemble/tests/balsam_tests/readme.balsam_tests.txt (hunk 8 above)
Comment (shuds13) [reply libensemble/tests/balsam_tests/readme.balsam_tests.txt]: Same

File: libensemble/executors/executor.py [hunk 9]
```diff
@@ -1,11 +1,11 @@
 """
 This module contains a
-``job_controller`` and ``job``. The class ``JobController`` is a base class and not
+``executor`` and ``task``. The class ``Executor`` is a base class and not
```
shuds13: I guess it should be an executor rather than a executor.

libensemble/executors/mpi_executor.py
```diff
@@ -1,38 +1,38 @@
 """
-This module launches and controls the running of MPI jobs.
+This module launches and controls the running of MPI tasks.
```
shuds13: This raises a concern over using "tasks". As MPI tasks refers usually to MPI ranks so this could be confusing. I would say we change this to MPI applications.

shuds13:
We should probably stick to camel case convention for Class names to be consistent. I can see using underscores makes a look a bit clearer - any other opinions on this? Is it worth breaking convention?

class MPI_Executor(Executor):
v
class MPIExecutor(Executor):

class Balsam_MPI_Executor(MPI_Executor):
v
class BalsamMPIExecutor(MPIExecutor):

docs/overview_usecases.rst
jlnav: "Applications" is a good word for describing to users and within the docs exactly what the executor is responsible for. I'm checking myself, but if anyone finds somewhere where "jobs", "tasks", or "scripts" should be "applications", please let me know.

docs/tutorials/executor_forces_tutorial.rst
jlnav: I think it's tasks now. Does this "Executor Variants" paragraph still make sense as a whole?

jlnav:
> We should probably stick to camel case convention for Class names to be consistent. I can see using underscores makes a look a bit clearer - any other opinions on this? Is it worth breaking convention?
> 
> class MPI_Executor(Executor):
> v
> class MPIExecutor(Executor):
> 
> class Balsam_MPI_Executor(MPI_Executor):
> v
> class BalsamMPIExecutor(MPIExecutor):

I went back and forth about this when changing the class names. The cleanliness issue comes from one of the components (MPI) of the camel case names being all upper-case.

How about `BalsamMpiExecutor(MpiExecutor)` as a possibility? If not, I can set the class names to traditional camel case.

shuds13:
I think we should stick with camel case convention - from what I see elsewhere its used and I think its still readable.

https://stackoverflow.com/questions/2853531/how-do-you-pep-8-name-a-class-whose-name-is-an-acronym

shuds13:
I think I would also leave launcher.launch (as this is internal to the Executor and I would say its directly launching jobs) - in launcher.py.

jmlarson1:
@jlnav Please run the following in the base `libensemble` directory and rename or remove files matching:
`$ find . -type f -iname "*job*"`

docs/platforms/platforms_index.rst
```diff
@@ -10,7 +10,7 @@ Across allocated compute nodes, libEnsemble's flexible architecture lends itself
 best to two general modes of worker distributions. The first mode we refer
 to as **centralized** mode, where the libEnsemble manager and worker processes
 are grouped on one or more nodes but through the libEnsemble executor or a
-job-launch command can execute calculations on the other allocated nodes:
+task-launch command can execute calculations on the other allocated nodes:
```
jlnav: I think I prefer "job" here because the context now refers to commands like `srun` or `qsub`, job-launch commands.

jmlarson1: I like the change to `srun`/`qsub`. Let's be specific, when possible.

jlnav:
> Let's try to have smaller PRs in the future :-)

Agreed. Each of the six changes listed on top could have justified it's own PR.

jmlarson1:
Live and learn

PR #346, opened 2020-01-31
Utils Reorganizing

jlnav:
This is a similar renaming/reorganizing branch to the other.

Functions for users to import directly within their calling scripts or functions are now within `tools.py`, renamed from `utils.py`. This means that calling scripts may contain:

`from libensemble.tools import parse_args, save_libE_output, add_unique_random_streams`

The ForkablePdb utility was moved to `tools.py`:

`from libensemble.tools import ForkablePdb`

Other utilities used _only_ by libEnsemble internally, like `launcher.py` and `timer.py` are within `libensemble/utils` (slight rename from `util`)


jlnav:
Heavily considering keeping user convenience functions under "utils" (this seems a universally understood term) and other modules like the Timer and Location Stack under "internal"

jmlarson1:
Either way is fine with me.

shuds13:
Another discussion point was  whether to make the user utilities/tools into a directory and possibly bringing gen/alloc/ support under the same directory.

E.g. Something like:

support/
  utils.py
  gen_utils.py
  alloc_utils.py

jlnav:
Update:

`from libensemble import libE` is now a valid import statement.

`from libensemble.executors import MPIExecutor` (or `Executor` or `BalsamMPIExecutor`) are now valid import statements.

Prior import statements that accomplish the above are _not_ broken.

There is now a `libensemble/tools` directory. Many of the user tools that were within `tools.py` before have been split into their own files, but are still accessible like:

`from libensemble.tools import parse_args`

 `alloc_funcs/support.py` and `gen_funcs/support.py` are now located within `libensemble/tools` as `alloc_support.py` and `gen_support.py` respectively.

The alloc and gen support functions can be imported like:
`from libensemble.tools import avail_worker_ids` or `from libensemble.tools.alloc_support import avail_worker_ids`

libensemble/__init__.py
```diff
@@ -9,3 +9,4 @@
 __credits__ = 'Argonne National Laboratory'
 
 from libensemble import libE_logger
+from .libE import libE, comms_abort
```
jmlarson1: I've never seen "from .package import name" with the dot. Is that intentional?

jlnav: Yes, this is intentional. It imports from libE in the same directory as the __init__

jlnav: This line makes ``from libensemble import libE`` possible.

shuds13:
I'm wondering if message_numbers should go in tools? It can be used by user scripts (e.g. sim/gen). 
Any opinions?

jlnav:
> I'm wondering if message_numbers should go in tools? It can be used by user scripts (e.g. sim/gen).
> Any opinions?

I agree. I'm considering making that another PR after this one is pulled in, since this one is rather large already.

jmlarson1:
> > I'm wondering if message_numbers should go in tools? It can be used by user scripts (e.g. sim/gen).
> > Any opinions?
> 
> I agree. I'm considering making that another PR after this one is pulled in, since this one is rather large already.

I think I disagree. Message_numbers aren't really tools. They are an essential part of tagging the different libEnsemble communications. That is, the tools seem to help users do common tasks so they don't have to write duplicate code. Message_numbers are our convention (admittedly, subject to change) for marking different messages.

jlnav:
> > > I'm wondering if message_numbers should go in tools? It can be used by user scripts (e.g. sim/gen).
> > > Any opinions?
> > 
> > 
> > I agree. I'm considering making that another PR after this one is pulled in, since this one is rather large already.
> 
> I think I disagree. Message_numbers aren't really tools. They are an essential part of tagging the different libEnsemble communications. That is, the tools seem to help users do common tasks so they don't have to write duplicate code. Message_numbers are our convention (admittedly, subject to change) for marking different messages.

That makes sense. ``message_numbers.py`` is similar to ``fields_keys.py`` that I concatenated together, which contains the allowed spec and libE fields. They both contain predefined constants used throughout the codebase. Should we create another directory, combine these files, leave them in ``/libensemble``?

shuds13:
OK, I think will can pull this. We can discuss further restructuring in future PR

PR #361, opened 2020-02-24
Fix/multisim dir label

jlnav:
Calculation directory naming routine for sims can now handle any combination of ``H_rows`` standalone rows or sequences of rows.

This means the first simulation working directory when running ``test_6-hump_camel_with_different_nodes_uniform_sample.py`` should be named ``sim0-4_worker1`` instead.

This PR also solves a small issue with that test, adds comments for the working directory system throughout, and adds some unit tests for many of the working directory functions.

Thoughts on the manager possibly creating the (empty) ``sim_input_dir`` automatically in the current working directory if it doesn't exist initially?


jmlarson1:
> Thoughts on the manager possibly creating the (empty) `sim_input_dir` automatically in the current working directory if it doesn't exist initially?

If it doesn't exist, shouldn't that be an error?

jlnav:
> If it doesn't exist, shouldn't that be an error?

This currently is an error, but for many of our regression tests and Forces, an empty ``sim_input_dir`` is required. If the user doesn't have any input files, but still wants a per-calculation output directory structure, maybe they shouldn't need to create an empty directory first.

jmlarson1:
> > If it doesn't exist, shouldn't that be an error?
> 
> This currently is an error, but for many of our regression tests and Forces, an empty `sim_input_dir` is required. If the user doesn't have any input files, but still wants a per-calculation output directory structure, maybe they shouldn't need to create an empty directory first.

I agree if someone wants per-calc directories, we shouldn't require them to make an empty directory first. But I think that's different than making an empty `sim_input_dir`, no?

shuds13:
Well we set sim_input_dir to an existing path. I can see if might be better to have an option
to just create empty dirs in place. In essence then you just giving a name rather than
a path. So how do we specify that?

shuds13:
That said, I suspect for most user cases, there will be input, so this will be an unusual case except in testing.

jlnav:
> I agree if someone wants per-calc directories, we shouldn't require them to make an empty directory first. But I think that's different than making an empty `sim_input_dir`, no?

Sorry for the confusion. All per-calc directories are copies of the directory specified by ``'sim_input_dir'``. To use per-calc directories, ``libE_specs['sim_input_dir']`` must be defined to some already-existing directory. 

In the case of Forces, there aren't any input files to copy over, so the directory can be empty. But right now it must exist! The user must remember to make it first before calling libE.

The solution could simply be if ``libE_specs['sim_input_dir']`` refers to a directory that doesn't exist, then libE makes that directory.

shuds13:
My problem with that is if the user has got a directory but specified the path wrong. They may get unexpected behaviour

jmlarson1:
> My problem with that is if the user has got a directory but specified the path wrong. They may get unexpected behaviour

Exactly. I think the most-likely case is that a directory not being there means it was mistyped. So I think an error should be raised. (Otherwise, if we make it and carry on, then there will likely be other issues. "Why can't my calculation find that file! It's clearly in this directory!"

jlnav:
> My problem with that is if the user has got a directory but specified the path wrong. They may get unexpected behaviour

Hmm, you're right. There's also the circumstance where setting ``libE_specs['sim_input_dir'] = '.'`` or another wrong path could produce irritating results.

docs/tutorials/local_sine_tutorial.rst
```diff
@@ -302,8 +300,9 @@ circumstances where Python's multiprocessing does not. In this section, we'll
 explore modifying the above code to use MPI instead of multiprocessing.
 
 We recommend MPICH_ for this tutorial, which can be found for a variety of systems
-here_. You also need mpi4py, which can be downloaded via ``pip3 install mpi4py``.
-If this doesn't work, try appending ``--user`` to the end of the command.
+here_. You also need ``mpi4py``, which can be downloaded via ``MPICC=$(which mpicc) pip3 install mpi4py``.
```
shuds13: I'm thinking, though need to check to be certain, that this line would do the default anyway. I have seen it recommended like this before, but I think mpi4py will grab this compiler. However, they might want to specify something different. On Bebop you have to use mpiicc to get actual intel compiler. On a cray system it might just be cc. I've been trying to find a line that always works. But maybe we should just explain it. eg.

can be downloaded via ``pip3 install mpi4py``

or if you are using a different MPI ``MPICC=<path/to/mpi C compiler>  pip3 install mpi4py``

see: https://mpi4py.readthedocs.io/en/stable/install.html
or maybe we just say if you are using a different MPI compiler see this link.

PR #362, opened 2020-02-25
Feature/renaming tests

jmlarson1:
Renaming regression tests to better convey what they test


jmlarson1:
Failing for a known issue

PR #369, opened 2020-02-27
Feature/linelength 120

jmlarson1:
Setting a linelength


jmlarson1:
> I haven't had any success yet.

Great. (Don't spend time on figuring out something this trivial. I've reverted that file.)

PR #393, opened 2020-03-13
Feature/improve dup comms

shuds13:
Use context management to free libE communicators.

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -235,7 +235,7 @@ def libE_mpi(sim_specs, gen_specs, exit_criteria,
 
         # Worker returns a subset of MPI output
         libE_mpi_worker(comm, sim_specs, gen_specs, libE_specs)
-        return [], persis_info, []
+        return [], {}, []
 
 
 def libE_mpi_manager(mpi_comm, sim_specs, gen_specs, exit_criteria, persis_info,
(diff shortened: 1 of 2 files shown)


libensemble/libE.py
```diff
@@ -204,31 +216,26 @@ def libE_mpi(sim_specs, gen_specs, exit_criteria,
...
+            libE_nodes = set(comm.allgather(local_host))
+            exctr.add_comm_info(libE_nodes=libE_nodes, serial_setup=is_master)
+
+        # Run manager or worker code, depending
+        if is_master:
+            return libE_mpi_manager(comm, sim_specs, gen_specs, exit_criteria,
+                                    persis_info, alloc_specs, libE_specs, H0)
+
         # Worker returns a subset of MPI output
         libE_mpi_worker(comm, sim_specs, gen_specs, libE_specs)
```
jmlarson1: Is `worker_main` not returning anything? Can the `persis_info` in the next line be updated from `worker_main`?

jlnav: ``worker_main`` in it's current form just initiates worker instances. ``persis_info`` is first communicated to a worker in a unit of work through the communicator, and back to the manager through the communicator as well, in a loop initialized within ``worker_main`` but without anything returned to it.

shuds13: Yes, the worker does not return anything. It is merely to satisfy the interface. Perhaps it should return an empty dictionary?

jmlarson1: Sounds good

Issue #382, opened 2020-03-05
Check sim['in'] are in gen['out']


shuds13:
I propose check_inputs() should check that if a gen exists, then anything in sim['in'] is defined in gen['out].

Are there possible exceptions to this?

jmlarson1:
I can think of cases where everything into the sim is not in `gen_specs['out']`. Perhaps resources to be used is an `alloc_spec['out']`. Perhaps the some filename or mesh information is computed beforehand and given in `H0`. Is it enough to do a check that everything in `sim_spec['in']` is a key in `H` (after it has been created)?

jmlarson1:
Addressed by #395

PR #395, opened 2020-03-16
Adding tests for various 'in' fields

jmlarson1:
In response to #382

Checks that 'in' fields are in one of:
- libE_fileds
- sim_specs['out']
- gen_specs['out']
- alloc_specs['out']
- H0

### DIFF
--- a/libensemble/tests/unit_tests/test_libE_main.py
+++ b/libensemble/tests/unit_tests/test_libE_main.py
@@ -156,7 +156,7 @@ def check_assertion(libE_specs, alloc_specs, sim_specs, gen_specs, exit_criteria
 def test_checking_inputs_noworkers():
     # Don't take more points than there is space in history.
     sim_specs, gen_specs, exit_criteria = setup.make_criteria_and_specs_0()
-    H0 = {}
+    H0 = np.empty(0)
 
     # Should fail because only got a manager
     libE_specs = {'comm': MPI.COMM_WORLD, 'comms': 'mpi'}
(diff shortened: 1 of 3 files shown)


libensemble/tests/unit_tests/test_libE_main.py
```diff
@@ -226,7 +226,12 @@ def test_checking_inputs_single():
 
     check_inputs(libE_specs=libE_specs)
     check_inputs(alloc_specs=alloc_specs)
-    check_inputs(sim_specs=sim_specs)
+    try:
+        check_inputs(sim_specs=sim_specs)
+    except AssertionError:
+        assert 1, "Fails because sim_specs['in']=['x_on_cube'] and that's not an 'out' of anything"
+    else:
+        assert 0, "Should have failed"
```
jlnav: Is there a possibility a user may want to perform a self-check of only their sim_specs for consistency, correct fields, etc. ? as in, like in the test, ``check_inputs(sim_specs=sim_specs)``. In which case I think this causes "valid" sim_specs to fail. Maybe I just don't understand the change, but to me it looks like if there's only one dictionary then it checks that the 'in' and 'out' fields match, which isn't a realistic expectation for a single dictionary. Thoughts?

jmlarson1: I see what you are saying, but issue #382 requested that `check_inputs` checked `sim_specs['in']`against other `['out']` fields. So I think that resolving the issue would break such a usage of `check_inputs` for single sim_specs checks.

Thoughts @shuds13 ?

jlnav: That's fair. Since a user has to at least define both ``sim_specs`` and ``gen_specs`` anyway, it's reasonable to expect they'd want to check both simultaneously for consistency, and also (with this PR) interoperability. Thanks!

PR #399, opened 2020-03-18
Debugging/hanging persis aposmm

jmlarson1:


shuds13:
Still need to talk about approach to dealing with subprocesses. If we explicitly terminate them, then PETSc calls mpi_abort.  See last commit.

shuds13:
Ok, I've undone the process terminate block. Think this is ready to merge.

PR #400, opened 2020-03-20
Debugging/persis aposmm comms test

jmlarson1:
Increasing the comms testing of persistent aposmm. `local` comms doesn't work with the external localopt regression test, perhaps something in the `subprocess` call.


jlnav:
When I commented out lines 40 and 41 in ``test_persistent_aposmm_external_localopt.py`` that cause the test to exit if using local and ran the test with multiprocessing, it appeared to work fine on my laptop.

jmlarson1:
I was not checking correctly. Thanks for the silly catch. (Both TCP and local work for me locally.)

PR #402, opened 2020-03-20
Adding a test to cover persis_aposmm exception raising

jmlarson1:
Is it possible to test the exception raising in `persis_aposmm`?

This branch has a new test `(libensemble/tests/regression_tests/test_persistent_aposmm_exception.py`) that raises the exception, but it aborts.

### DIFF
--- a/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
+++ b/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
@@ -35,6 +35,7 @@ def assertion(passed):
         else:
             MPI.COMM_WORLD.Abort(1)  # Abort with failure
     else:
+        print("\n\nCorrect NLopt exception caught anyway.")
         assert passed
 
 
--- a/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
+++ b/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
@@ -35,8 +35,9 @@ def assertion(passed):
         else:
             MPI.COMM_WORLD.Abort(1)  # Abort with failure
     else:
-        print("\n\nCorrect NLopt exception caught anyway.")
         assert passed
+        print("\n\nCorrect NLopt exception caught anyway.")
+
 
 
 nworkers, is_master, libE_specs, _ = parse_args()
--- a/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
+++ b/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
@@ -36,8 +36,7 @@ def assertion(passed):
             MPI.COMM_WORLD.Abort(1)  # Abort with failure
     else:
         assert passed
-        print("\n\nCorrect NLopt exception caught anyway.")
-
+        print("\n\nException received as expected")
 
 
 nworkers, is_master, libE_specs, _ = parse_args()
--- a/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
+++ b/libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
@@ -25,6 +25,18 @@ from libensemble.alloc_funcs.persistent_aposmm_alloc import persistent_aposmm_al
 from libensemble.tools import parse_args, add_unique_random_streams
 
 
+def assertion(passed):
+    """Raise assertion or MPI Abort"""
+    if libE_specs['comms'] == 'mpi':
+        from mpi4py import MPI
+        if passed:
+            MPI.COMM_WORLD.Abort(0)  # Abort with success
+        else:
+            MPI.COMM_WORLD.Abort(1)  # Abort with failure
+    else:
+        assert passed
+
+
 nworkers, is_master, libE_specs, _ = parse_args()
 
 if nworkers < 2:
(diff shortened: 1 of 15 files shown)


jmlarson1:
@shuds13 Any idea why this hangs on Travis? I thought it should fail.

jlnav:
> @shuds13 Any idea why this hangs on Travis? I thought it should fail.

I've been investigating this for a bit. The exception is caught by the manager and libEnsemble exits normally when running with MPI, but when running with multiprocessing the hang is actually within ``kill_proc_team()`` in libE, then ``result()`` in ``comms.py``

shuds13:
Multiprocessing:
If I do self.process.terminate() before the join that works OK for me, on my laptop, but not on travis.

PETSc will mpi_abort when its killed but that does not seem to affect the manager on my laptop (despite
mpi4py/petsc being imported at the start).

jmlarson1:
With regards to the multiprocessing, it seems to have worked perfectly on Travis:
https://travis-ci.org/github/Libensemble/libensemble/builds/666652208?utm_medium=notification&utm_source=github_status

We know PETSc might be an issue in general, but it shouldn't affect this NLopt test, right? 

The `push` build on Travis finished, but the `pr` build didn't for 3.5 and 3.7. In these cases, something happens with the `test_uniform_sampling_then_persistent_localopt_runs.py`, which is maintaining a single, persistent (nlopt) local optimization run that communicates directly with the manager. (The run is not a child processes.)

Perhaps the `try-except` starting line 94 of `libensemble/gen_funcs/uniform_or_localopt.py` (to catch the exception raised when receiving a `STOP_TAG` in line 61) are not how we are expecting a persistent worker to exit?

shuds13:
It does seem to be working. I want to check what happens with tests that used to fail on Travis. And will need to check other systems.

As for PETSc, it still gets imported (when aposmm is imported), so I think any time its given a terminate signal it will MPI Abort. The question is whether it only aborts the current (rank 0) process or effects the others also.I dont think mpi abort behaviour is defined in this situation where you may have multiple processes with rank 0. 

Also, the terminate() I added was to test if it works, but it does not make sense with the surrounding code. We could could always terminate or we could add a default timeout. There should not be a need to wait for any significant time here, as final_receive_and_kill() will have already waiting on any worker returns we want to wait for.

libensemble/tests/regression_tests/test_persistent_aposmm_exception.py
```diff
@@ -0,0 +1,86 @@
...
+    """Raise assertion or MPI Abort"""
+    if libE_specs['comms'] == 'mpi':
+        from mpi4py import MPI
+        if passed:
+            print("\n\nMPI will be aborted as planned\n\n", flush=True)
+            MPI.COMM_WORLD.Abort(0)  # Abort with success
+        else:
+            MPI.COMM_WORLD.Abort(1)  # Abort with failure
+    else:
+        assert passed
```
shuds13: What does the 'anyway' mean?

jlnav: I added that, partially to trigger a build and partially as a reminder that the PETSC ERROR is expected.

jlnav: "Even though the PETSC ERROR is displayed, the correct exception was caught from aposmm by libensemble" I guess was my thought process.

shuds13: Maybe say "Expected exception received' or something along those lines.

jlnav:
> We know PETSc might be an issue in general, but it shouldn't affect this NLopt test, right?

Could the ``from petsc4py import PETSc`` import be moved from file-level to ``run_local_tao()``? I was justifiably worried about speed and wasteful statements until I read https://stackoverflow.com/a/3095167. Thoughts?

> We could could always terminate or we could add a default timeout.

Local routines in libE use an ``on_cleanup()`` with an optional timeout, but maybe it could also use an ``on_abort()`` to terminate always.

shuds13:
Our previous experience was that importing PETSc anywhere else caused it to hang on import. I would like to investigate this more. I would much rather these imports were not at the top.

jmlarson1:
I agree with not having them at the top, but doing so caused slowdowns or errors, right?

jmlarson1:
I agree that we need more debugging

For example, looking at the output here:
https://travis-ci.org/github/Libensemble/libensemble/jobs/666966484?utm_medium=notification&utm_source=github_status
there are PETSc errors raised during the 
`test_persistent_aposmm_dfols`test, which should have nothing to do with PETSc.

jlnav:
I experimented a little with moving the PETSc import and running ``test_persistent_aposmm_tao.py``. Right now, with local comms, moving the import made the test take about forty seconds instead of about eight. With MPI, the test hangs. By itself, running ``from petsc4py import PETSc`` takes half a second.

shuds13:
> there are PETSc errors raised during the
> `test_persistent_aposmm_dfols`test, which should have nothing to do with PETSc.

As long as PETSc is imported you will get that PETSc error if you kill the process. This is caused if we kill the process before it has exited. When we close things down (we have already got all our results back), we now wait for a timeout period for it to finish, and if it has not we terminate the process (to prevent that hang on exception). 

We could play with the timeout so we give the process more time to close out, I only set it to one second. I think we should pull this into develop, and I will test on the HPC platforms.

PR #409, opened 2020-03-26
Debugging/persis aposmm return info

jmlarson1:
Adjusting the final receive in manager to get the `persis_info` out of the run.


shuds13:
Attempted solution (currently failing):
In _final_receive_and_kill a peris_stop is sent out to each active persistent worker*, which
is put in a handshake_pending list.
While the return of that signal is pending, the worker will not be marked inactive when receiving data.
When the returned signal is received by the manager the handshake_pending is removed so the
final receive of calculation data marks the worker as inactive.

What this is getting back is not what it would expect from a normal gen out, but
rather I think all H row. What do we want to get back here? Is it just persis_info?

*Is it possible to have an inactive persistent worker?

jmlarson1:
> *Is it possible to have an inactive persistent worker?

Yes. If a persistent worker has returned some information (e.g., points it requests to be evaluated) and is waiting (e.g., for their sim output value) then their `persis_state > 0` but `active = 0`.

shuds13:
The documentation should include the 'worker_id' field right?

jmlarson1:
Yes.

shuds13:
Also we will need to do the handshake with all persistent then, not just persistent active.

shuds13:
I have tested this solution when persistent generator is active and inactive. Another question is what we do on a timeout or Worker Exception. Currently these scenarios will not wait.

jmlarson1:
In a timeout, if the calculation is waiting to return, it would be nice to do the same "send PERSIS_STOP to calculation and blocking receive from worker (because we know the worker is going to be returning to the worker)"

Or is there something preventing the same operations from happening?

shuds13:
Nothing to prevent them from happening. The only consideration, and the reason we currently don't wait for workers on a timeout, is in case there is a long running simulation. I think for this case we should wait.

shuds13:
While we're here, the persis_info dump is called libE_history_at_abort_*.pickle

Would it be better named something like libE_persis_info_*.pickle

jmlarson1:
> Would it be better named something like libE_persis_info_*.pickle

Good idea. This has been included in PR #417

shuds13:
I would hold off on this branch for now. I think I'm close to having the alternative solution ready. Just checking a few things.

jmlarson1:
I'm fine with holding off. 

I want to note that we have an [occasional failure with terminating active persistent workers that needs to be addressed](https://travis-ci.org/github/Libensemble/libensemble/builds/669504179?utm_medium=notification&utm_source=github_status)

shuds13:
This was merged in via debugging/persis_aposmm_return_info_soln2

PR #413, opened 2020-03-27
Forces exception tests

jlnav:
Adding two (for now) simple and optional fail conditions within Forces. This will help test libEnsemble's handling of failures and exceptions in a variety of configurations across systems.

The first failure is a simple raised exception as the sim_f is initiated. This exception is caught and both log files are checked for the proper messages.

The second causes the forces task submission to fail. No exceptions are raised, but the Executor (correctly) retries up to four times. libE_stats.txt is checked.

These changes will get included in the templater repo and I'll test them on Bebop/Theta.


jlnav:
Bebop

- [ ] MPI - fail_on_sim

- [ ] MPI - fail_on_submit

- [ ] mproc - fail_on_sim

- [ ] mproc - fail_on_submit

Theta

- [ ] MPI/Balsam - fail_on_sim

- [ ] MPI/Balsam - fail_on_submit

- [ ] mproc/MOM - fail_on_sim

- [ ] mproc/MOM - fail_on_submit

- [ ] Big run - fail_on_sim

PR #415, opened 2020-03-31
Add missing import libensemble.utils in local sine tutorial

MaxThevenet:
The Simple Local Sine Tutorial is great and very smooth! In file `calling_script.py` though, `add_unique_random_streams` is used without explicit import of the required module. This PR fixes this.

### DIFF
--- a/docs/tutorials/local_sine_tutorial.rst
+++ b/docs/tutorials/local_sine_tutorial.rst
@@ -175,6 +175,7 @@ multiprocessing.
     from libensemble.libE import libE
     from generator import gen_random_sample
     from simulator import sim_find_sine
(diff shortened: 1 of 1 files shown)


jlnav:
Good catch! Thank you for fixing this and for feedback about the tutorial.

We recently reorganized the project so the ``add_unique_random_streams`` functions is actually within the ``libensemble.tools`` module, although this is subject to change. Sorry for the confusion. The full import line should be ``from libensemble.tools import add_unique_random_streams``

We also have an additional completed tutorial on our develop branch that introduces using libEnsemble to launch and monitor applications:

https://libensemble.readthedocs.io/en/develop/tutorials/executor_forces_tutorial.html

PR #418, opened 2020-04-01
Debugging/persis aposmm return info soln2

shuds13:
Alternative persistent close-down method

Only implemented for persistent_aposmm in first instance but  persistent gens could simply return FINISHED_PERSISTENT_GEN_TAG


jmlarson1:
Great. This looks good to pull in. Before doing so, I wanted to note that lines 24-27 of `executor_hworld.py` and lines 372, 431, and 458 `libE_worker.py` are no longer covered.

- As they both have to do with `MAN_SIGNAL_FINISH`, and we've changed our shutdown, is that code out of date?
- Or can we add a test (at a later date is fine) to get them covered again?

shuds13:
I don't think that test should be affected. I will take a look

shuds13:
I have fixed that issue. I'm happy to pull this in.

PR #420, opened 2020-04-06
Experimental/aposmm rc

shuds13:
A provisional method to prevent unnecessary imports in persistent_aposmm. By specifying the optimizer as a configuration option, other optimizers do not need to be imported. This also means mpi4py is not imported in APOSMM unless it is being used.

It would be interesting to see if this helps if there are issues with OpenMPI.


shuds13:
Note - this is a pull request for warpx branch.

PR #423, opened 2020-04-10
Copy or symlink exact lists of files into sim_dirs instead of copying a directory

### LINKED ISSUE #419: Be able to specify exactly what files to copy/symlink into sim_dirs, independent of any single sim_input_dir
Overall this may be a more natural and customizable approach to simulation directories. Currently, a single given directory and its contents is given to be copied for each sim, with options to exclude or symlink certain files out of that directory into a sim_dir. 

An alternative, and probably better approach, is the user provides explicit lists of files to copy or symlink into sim_dirs. These filepaths should naturally be independent instead of relative to any one given input directory, as is currently the case.

Until this is completed, a possible workaround is setting the sim_input_dir to the one parent directory containing all directories with wanted inputs (there's technically always one!), then being explicit with possibly long filepaths in lists provided to copy_input_files and symlink_input_files.

jlnav:
These changes should simplify the interface for simulation directories. 

To use these features, users no longer have to place all their input files into a certain directory and include/exclude files in it; they can now simply provide lists of filepaths to copy or symlink into each simulation directory.

Using sim_dirs without any inputs only requires ``libE_specs['make_sim_dirs'] = True``. An empty directory is no longer needed to initialize tests like Forces.

Most related settings have had their names changed to better communicate their compatibility. Some settings like ``'copy_input_to_parent'`` and ``'clean_ensemble_dirs'`` have been removed.

Closes #419


jlnav:
- [ ]  Fix/evaluate better_copytree

- [ ]  Fix copy_back not being by-worker when ensemble directory by-worker.

jmlarson1:
Are we testing `'use_worker_dirs'`? 
https://coveralls.io/builds/30003017/source?filename=libensemble%2FlibE_worker.py#L257

jlnav:
> Are we testing `'use_worker_dirs'`?
> https://coveralls.io/builds/30003017/source?filename=libensemble%2FlibE_worker.py#L257

Good catch. I guess this proves that intermediate directory switching isn't necessary if a sim_dir has already been registered and can be switched to, since the output directory structures are passing tests.

Thank you for the help!

jlnav:
- [ ] Fix ``'use_worker_dirs'`` and intermediate switching

- [ ] Check ``_make_calc_dir()`` for redundant operations (some re-calculated values should be worker attributes)

jmlarson1:
There is also some loc_stack lines that were uncovered. Not sure if they are needed:
https://coveralls.io/builds/30003017/source?filename=libensemble%2Futils%2Floc_stack.py#L23

jlnav:
> There is also some loc_stack lines that were uncovered. Not sure if they are needed:
> https://coveralls.io/builds/30003017/source?filename=libensemble%2Futils%2Floc_stack.py#L23

Fair enough, the ``except FileExistsError`` statements aren't needed if libEnsemble's expected behavior is to crash if it starts reusing a preexisting ensemble directory. The first statement is covered because the second iteration of ``test_sim_dirs.py`` reuses the ensemble directory from the first iteration. I assume libEnsemble should actually be prevented from doing this. 

Would you prefer libEnsemble to exit and warn the user if it starts reusing a previously existing ensemble directory, or automatically make a new unique directory? 

Any thoughts related to this or how previous output should be treated?

jlnav:
> @shuds13 is this the correct usage?

I think I did something flakey with git at some point that required this. I couldn't merge a change on gen_support in develop onto my feature branch.

jlnav:
> Are you branching off master?

This is off develop. I tried rebasing with develop, but only received "up-to-date" messages even though there was a blatant difference between gen_support on develop and on this branch. In any case, even with this commit there is no gen_support diff listed under Files changed, so I think this gen_support version now matches develop?

jlnav:
- [ ] Diagnose instability on Travis (including directories made with "w3"????)

shuds13:
My suggestions for naming of options and docs:

        'sim_dirs_make' [boolean] :
            Whether to make simulation-specific directories for each sim call.
            This will create a sim dir for each simulation, even if no sim_input_dir is specified.
            Default: False
        'ensemble_dir' [string] :
            Path to main ensemble directory containing calculation (sim/gen) directories. Default: './ensemble'
            This is only created if calculation directories are created.
        'use_worker_dirs' [boolean] :
            Whether to organize calculation (sim/gen) dirs under worker directories.
            Default: False
        'sim_dir_copy_files' [list] :
            List of paths to files to copy into each sim dir.
        'sim_dir_symlink_files' [list] :
            List of paths to files to symlink into each sim dir.
        'ensemble_copy_back' [boolean] :
            Whether to copy back directories within ensemble_dir back to launch location.
            This option is useful if the ensemble_dir is placed on worker local storage.
            Default: False [*I WOULD MAKE THE DEFAULT TRUE]
        'sim_input_dir' [string] :
            Copy this directory and it's contents for each simulation-specific directory.

When have gen_dirs we may add:

        'gen_dirs_make' [boolean] :
            Whether to make generator-specific directories for each gen call.
            Default: False
        'gen_dir_copy_files' [list] :
            List of paths to files to copy into each gen dir.
        'gen_dir_symlink_files' [list] :
            List of paths to files to symlink into each gen dir.
        'gen_input_dir' [string] :
            Copy this directory and it's contents for each generator-specific directory.

Please argue...

shuds13:
Questions (I've used the current option naming in these questions):
- If sim_dirs_make is not set does this mean sim_dir_copy_files,sim_dir_symlink_files are ignored or set implicitly on?
- If sim_dirs_make but sim_input_dir is set, presumably the sim_input_dir overrides?
- If sim_input_dir is set, and then a sim_dir_symlink_files specifies on of those files - does it override and craete symlink instead of copy?
- is copy_input_to_parent removed?
- Does sim_dir_copy_files/sim_dir_symlink_files work on directories? Do they work with wildcards (*)?

I have kept the order of options above for comparison, but I would probably order them so sim_dir* specific are together, or are we are going with alphebetical order for all options?

jlnav:
> If sim_dirs_make is not set does this mean sim_dir_copy_files,sim_dir_symlink_files are ignored or set implicitly on?

They are currently ignored. Current pseudocode-logic in the worker is:
```
if either 'sim_dirs_make' or 'sim_input_dir' in libE_specs:
   do calc with sim_dirs
```

I could replace this with:

```
if any calc_dir settings are in libE_specs:
   do calc with sim_dirs
```

This should work since each setting has compatible default values if not specified. For example, with this change, only setting ``libE_specs['sim_dir_symlink_files'] = './config.txt'`` will create calc_dirs in ``./ensemble``, not organized under worker-dirs, with the above file symlinked in.


> If sim_dirs_make but sim_input_dir is set, presumably the sim_input_dir overrides?

sim_dirs_make was originally intended as an on/off switch for these features, but as discussed above can probably be removed. To answer specifically: if either is set, calc_dirs are created. So it would override 'sim_dirs_make' = False if that was set for some reason.

> If sim_input_dir is set, and then a sim_dir_symlink_files specifies on of those files - does it override and craete symlink instead of copy?

Testing this just reminded me that I had already used an assertion to check there aren't collisions between files to symlink and copy. Without this assertion, os.symlink raises FileExistsError if there's a collision, since copying happens before.

Out of the options of "enforce separation via assertion", "copying overriding symlinking", and "symlinking overriding copying", which is preferable? I'm partial to the first.

> Is copy_input_to_parent removed?

Yes. It was a bit complicated to both implement and describe/justify in the docs, and I haven't seen any use-cases yet where the feature would prove useful. If they're there, please say so!

> Does sim_dir_copy_files/sim_dir_symlink_files work on directories? Do they work with wildcards (*)?

Those options work on directories. Wildcards current don't work, but I could try something with ``glob``.

> I have kept the order of options above for comparison, but I would probably order them so sim_dir* specific are together, or are we are going with alphebetical order for all options?

I agree with grouping related sets of options together in the docs.

shuds13:
When you say sim_dirs_make could be removed, I presume you still want it for case of creating empty directories. 

As you say, they can use glob to list files with a wildcard, so I think thats fine. The glob will create a list so we dont need to deal with that.

I am fine for copy to override sym-linking.

I'm Ok for copy_input_to_parent to be removed for now and see if a use-case arises. The idea was for a scenario where you had worker local storage where you want to keep one copy. But we really dont do that currently.

shuds13:
Also, from my name changes comment, I would probably make the copy back from scratch True by default.

jlnav:
- [ ] ``'ensemble_dir'`` without ``'sim_dirs_make'`` makes only ensemble directory with workers switching there.

shuds13:
In loc_stack.py
Have you considered something like:
src_path = os.path.abspath(os.path.expanduser(os.path.expandvars(file_path)))
to pick up use of $HOME and ~/
?

shuds13:
Should this combo work?
libE_specs['sim_dirs_make'] = False
libE_specs['ensemble_dir_path'] = w_ensemble
libE_specs['use_worker_dirs'] = True

jmlarson1:
Should we be testing the `FileExistsError`? 

https://coveralls.io/builds/30471193/source?filename=libensemble%2FlibE_worker.py#L270
or 
https://coveralls.io/builds/30471193/source?filename=libensemble%2Futils%2Floc_stack.py#L41

jlnav:
> Should we be testing the `FileExistsError`?
> 
> https://coveralls.io/builds/30471193/source?filename=libensemble%2FlibE_worker.py#L270
> or
> https://coveralls.io/builds/30471193/source?filename=libensemble%2Futils%2Floc_stack.py#L41

Yes. I'm working on some unit tests for those cases.

PR #425, opened 2020-04-13
Experimental/warp x

jmlarson1:


jmlarson1:
@shuds13 In your opinion, should we change all `persistent_aposmm` regression tests to use `rc.aposmm_optimizer`? Or is there reason to test both with and without using `rc.aposmm_optimizer`? 

It is not currently [being tested in our regression tests.](https://coveralls.io/builds/30055355/source?filename=libensemble%2Fgen_funcs%2Fpersistent_aposmm.py#L20)

shuds13:
Yes I think we should add to tests. If its not included it will just import everything. I don't think its essential to test that as well.

jmlarson1:
> Yes I think we should add to tests. If its not included it will just import everything. I don't think its essential to test that as well.

How do I re-set the optimizer from `nlopt` in the first call to libE in the script to `scipy` in the second call to libE in the script?

shuds13:
> > Yes I think we should add to tests. If its not included it will just import everything. I don't think its essential to test that as well.
> 
> How do I re-set the optimizer from `nlopt` in the first call to libE in the script to `scipy` in the second call to libE in the script?

Thats a good point. It may be better for me to make it a list rather than specifying one optimizer.  If it was just scipy we could maybe just always import that - in fact we import scipy functions anyway. But a list might be a better general solution.

jmlarson1:
> > > Yes I think we should add to tests. If its not included it will just import everything. I don't think its essential to test that as well.
> > 
> > 
> > How do I re-set the optimizer from `nlopt` in the first call to libE in the script to `scipy` in the second call to libE in the script?
> 
> Thats a good point. It may be better for me to make it a list rather than specifying one optimizer. If it was just scipy we could maybe just always import that - in fact we import scipy functions anyway. But a list might be a better general solution.

I agree. Can you put it on the apossm_rc branch?

shuds13:
Yes, which test uses both?

jmlarson1:
`test_persistent_aposmm_periodic.py`

shuds13:
I've pushed the required fix. Do you want to change the name to aposmm_optimizers?  I'll let you decide what to call it.

By the way its implemented for non-persistent aposmm also.

shuds13:
Seems that subprocesses are failing on OSX, has anyone had this before?

jlnav:
> Seems that subprocesses are failing on OSX, has anyone had this before?

I've been looking into it today. conda started installing an out-of-date version of clang compared to previously successful builds. For OSX, version 9.0.1 is gone with only version 4.0.1 available. I'm looking into alternatives from conda-forge.

shuds13:
As the OSX problem does not seem specific to this branch. any objections to pulling it in?

jmlarson1:
Let's make sure the last merge was correct (it was non-trivial), but I'm fine pulling in even if OSX fails.

PR #426, opened 2020-04-14
Experimental/aposmm rc

shuds13:
Some fixes plus extend to aposmm as well as persistent_aposmm.


jmlarson1:
@shuds13
1.  I'm not sure which side of the conflicts to take for `scaling_tests/warpx` files. Do you?
2. Does this approach not work if the local optimizer in `persistent_aposmm` is changed between two libE calls in the same script? (I couldn't get `test_persistent_aposmm_periodic.py` to work with this new import approach.)

shuds13:
I'll take a look and resolve.

PR #428, opened 2020-04-16
Docs/more gens

jmlarson1:


shuds13:
The persistent_aposmm docs does not include the main aposmm function (its not in __all__). Is this deliberate?

shuds13:
I want to pull this in now and I will make sure we consolidate develop. I can pull into aposmm_rc and then pull that again into warpx and then pull that into develop.

I think we may then want to do a bit more work on this. My main issue is whether this is supposed to present a API to the user of the gens, or just example gens. Also for the APOSMM user, we want them to have the information they need for setting up sim/gen specs and what optimizer options, but not sure we want to provide internal APOSMM routines.

jmlarson1:
Sounds good

PR #432, opened 2020-04-17
Attempts to fix OSX Executor failures

jlnav:
keeping xcode version update.


jmlarson1:
Looks great. Thanks @jlnav !

PR #434, opened 2020-04-20
Support serial PETSc in APOSMM.

shuds13:
Remove MPI from APOSMM modules.
Replace MPI.COMM_SELF with PETSc.COMM_SELF


shuds13:
It should avoid them if you are using a serial build of PETSc. This should work with both serial and parallel builds.

jmlarson1:
I understand. The issue was with PETSc error raising when built with MPI, not when running petsc4py when mpi4py was also imported.

shuds13:
Yes, PETSc imports MPI when it is a parallel build.

PR #436, opened 2020-04-24
Conda installation instructions on README and HPC pages

jlnav:
With slight adjustments elsewhere. Please feel free to offer suggestions, or make modifications yourself.

Thank you!

### DIFF
--- a/docs/platforms/bebop.rst
+++ b/docs/platforms/bebop.rst
@@ -39,15 +39,15 @@ libEnsemble can then be installed via ``pip`` or ``conda``. To install via ``pip
 
     (my_env) user@login:~$ pip install libensemble
 
-Installing libEnsemble via ``conda`` may contain platform-specific
-optimizations for dependencies like NumPy. You can also choose which MPI libEnsemble uses:
+To install via ``conda``:
 
 .. code-block:: console
 
     (my_env) user@login:~$ conda config --add channels conda-forge
-    (my_env) user@login:~$ conda install -c conda-forge libensemble=*=mpi_mpich*    # with MPICH
-             OR            conda install -c conda-forge libensemble=*=mpi_openmpi*  # with Open MPI
-             OR            conda install -c conda-forge libensemble                 # without MPI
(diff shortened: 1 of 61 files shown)


docs/platforms/bebop.rst
```diff
@@ -25,15 +25,30 @@ Installing libEnsemble and Dependencies
...
 
     (my_env) user@login:~$ CC=mpiicc MPICC=mpiicc pip install mpi4py --no-binary mpi4py
+
+libEnsemble can then be installed via ``pip`` or ``conda``. To install via ``pip``:
+
+.. code-block:: console
+
     (my_env) user@login:~$ pip install libensemble
 
+Installing libEnsemble via ``conda`` may contain platform-specific
```
shuds13, line 42: Think I would remove this line. They are compiled for the general architecture but may not have very specific flags (like AVX-512), and not really different than a pip wheel (I think).

docs/platforms/bebop.rst
```diff
@@ -25,15 +25,30 @@ Installing libEnsemble and Dependencies
...
+optimizations for dependencies like NumPy. You can also choose which MPI libEnsemble uses:
+
+.. code-block:: console
```
shuds13, line 45: I would not recommend mpich/openmpi for bebop etc... These systems should just get the default (no-mpi) version. Note: It's true, if they know what they are doing on an mpich-derivative system, they could point to/copy the system MPI libs and it should work due to MPICH ABI compatibility, but I think thats too advanced/potentially confusing for docs (esp. as LD_LIBRARY_PATH won't work).

Your instructions for MPI variants should go in the standalone installation section (do we still have that? I dont want to clog up the quick start too much, we should also maybe list spack variants there, but at most include a link here.)

jlnav: Thanks for the suggestions. Which platforms do you mean by "bebop etc.", each HPC platform on the docs?

Our only dedicated installation section is part of quickstart / README, and then we have platform-specific lines in each platform guide. We should make an advanced installation section somewhere that includes conda and spack variants.

PR #440, opened 2020-05-04
Feature/restructure mpi executor

shuds13:
Major restructure of MPI executor.

- Create MPIRunner class with information for each type of MPI Runner and allow custom
- Move to method 2 for portable options with jsrun (num_procs -> resource sets)
- Parses extra_args for recognized no. procs/nodes/ppn options for auto-resources
- Add customizable system settings (custom_info option)

### DIFF
--- a/libensemble/comms/comms.py
+++ b/libensemble/comms/comms.py
@@ -105,9 +105,9 @@ class QComm(Comm):
     These can be used with threads or multiprocessing.
     """
 
-    # Integer count - shared amongst Processes in multiprocessing
+    # Integer count - shared amongst processes
     lock = Lock()
-    _ncomms = Value('i', 0)  # todo - check: Maybe this should be in QCommProcess
+    _ncomms = Value('i', 0)
 
     def __init__(self, inbox, outbox, copy_msg=False):
         "Set the inbox and outbox queues."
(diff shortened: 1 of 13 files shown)


jmlarson1:
> Looks good. Are `dry_run` and `extra_args` the new user-facing additions?

I think `test` became `dry_run`

jmlarson1:
@shuds13 Any reason for the 1.8% drop in coverage?

shuds13:
> @shuds13 Any reason for the 1.8% drop in coverage?

Partly, I can add some tests. Also some lines to check why not covered.

shuds13:
> Looks good. Are `dry_run` and `extra_args` the new user-facing additions?

extra_args is already there. But this will parse the node/procs configuratiom parameters.

API changes:
- dry_run was called test
- mpi_executor setup includes an option custom_info - I'll add a bit in the docs - lets you define system info which overrides auto-detected parameters.

shuds13:
I have added testing and found a couple of minor fixes in the process.

As there is a docs branch open, I will update the docs within that to avoid conflicts

shuds13:
I don't think either of the test failures are to do with the code. One is the timing due to virtualization on OSX, the other is an initialization error. @jmlarson1 are you OK if pull this in?

jlnav:
> I don't think either of the test failures are to do with the code. One is the timing due to virtualization on OSX, the other is an initialization error.

These pop up occasionally. If it's any comfort, I've never seen such timing errors occur on my OSX machine, only on Travis's VMs.

PR #441, opened 2020-05-05
Forces - Functions to test ensemble_dir, exception logging

jlnav:
Adds ``forces_support.py`` file to Forces test, containing functions to test exception logging (when relevant) and structure/output of ensemble directory.


shuds13:
Are you updating the forces.c in here also?

PR #453, opened 2020-05-12
Changes to comply with flake8 3.8.0

shuds13:


jmlarson1:
I am confused as to why the flake8-3.7.9 that I was running locally allowed the errors 
`E741 ambiguous variable name 'O'`.

But I'm happy this was easy to fix.

PR #454, opened 2020-05-13
Testing/separate by comms

jmlarson1:


jmlarson1:
There was one troublesome test where we had a 1-second timeout but the gen_f didn't get started in that time (on OSX). I think increasing it will solve the issue. 

I'll merge when it passes.

PR #456, opened 2020-05-15
Making a standalone persistent_aposmm example

jmlarson1:
Modifies `persistent_aposmm` to perform its own gradient and function evaluations when in standalone mode. To make this change minimally invasive, this is accomplished by having `persistent_aposmm` make the data structures that should be received from the manager.


shuds13:
Is this ready to pull in? I'm going to be making more minor fixes to docs, I notice this touches docs in a few places.

PR #473, opened 2020-06-03
Trying to kill existing processes at the end of the run instead of gi…

jmlarson1:
…ving them funciton values and hope they stop

### DIFF
--- a/libensemble/gen_funcs/aposmm_localopt_support.py
+++ b/libensemble/gen_funcs/aposmm_localopt_support.py
@@ -160,18 +160,19 @@ class LocalOptInterfacer(object):
 
     def destroy(self, previous_x):
 
-        while not isinstance(previous_x, ConvergedMsg):
-            self.parent_can_read.clear()
-            if self.grad0 is None:
-                self.comm_queue.put((previous_x, 0*np.ones_like(self.f0),))
-            else:
-                self.comm_queue.put((previous_x, 0*np.ones_like(self.f0), np.zeros_like(self.grad0)))
+        # while not isinstance(previous_x, ConvergedMsg):
+        #     self.parent_can_read.clear()
+        #     if self.grad0 is None:
+        #         self.comm_queue.put((previous_x, 0*np.ones_like(self.f0),))
+        #     else:
+        #         self.comm_queue.put((previous_x, 0*np.ones_like(self.f0), np.zeros_like(self.grad0)))
 
-            self.child_can_read.set()
-            self.parent_can_read.wait()
+        #     self.child_can_read.set()
+        #     self.parent_can_read.wait()
 
-            previous_x = self.comm_queue.get()
-        assert isinstance(previous_x, ConvergedMsg)
+        #     previous_x = self.comm_queue.get()
+        # assert isinstance(previous_x, ConvergedMsg)
+        self.process.kill()
         self.process.join()
         self.comm_queue.close()
         self.comm_queue.join_thread()


shuds13:
PETSc defaults to mpi_abort within the signal handler for SIGTERM. We can get around by using SIGKILL or SIGINT, but multiprocessing has a limited interface here. Kill works, but is only supported from Python 3.7. We can bypass interface like this:

    self.process._popen._send_signal(signal.SIGKILL)

or

    self.process._popen._send_signal(signal.SIGINT)
    
Otherwise we do not terminate for PETSc. We need a regression test for this, like test_persistent_aposmm_exception.py, but using PETSc.

There is an on-going discussion about PETSc calling MPI_ABORT in the SIGTERM signal handler   https://gitlab.com/petsc/petsc/-/merge_requests/2745

If we only do this on exiting APOSMM, then we may want to put `clean_up_and_stop` in a finally clause.

jmlarson1:
Is there any downside to adjusting the existing `terminate` to send a `sigkill`? Is that difficult to do?

shuds13:
> Is there any downside to adjusting the existing `terminate` to send a `sigkill`? Is that difficult to do?

Not quite sure what you mean, can you give an example?

The multiprocessing terminate function issues a signal.SIGTERM to the process. I could use as in the above, or we could bypass multiprocessing altogether and just do:

    os.kill(self.process.pid, signal.SIGKILL)

Then we're not using internals.

ps. the functions in multiprocessing are:

    def terminate(self):
        self._send_signal(signal.SIGTERM)

        
    def _send_signal(self, sig):
        if self.returncode is None:
            try:
                os.kill(self.pid, sig)
            except ProcessLookupError:
                pass
            except OSError:
                if self.wait(timeout=0.1) is None:
                    raise

jmlarson1:
I thought we could hijack `terminate` to send `sigkill` but given that `terminate` just sends a

    self._send_signal(signal.SIGTERM)

it doesn't seem like we can tell the signal to send. Can we just define/use our own function like 

```
def localopt_run_terminate(self):
    self._send_signal(signal.SIGKILL)
```
?

dbindel:
Recommended protocol for Unix-based system is to try to send a terminate request (SIGTERM), which can be caught by the application to do clean-up; and then send a kill signal (SIGKILL) if the application remains unresponsive after SIGTERM.  Things are a little more complicated yet because you sometimes want to kill a whole process group (the parent process + any child processes it started).  Things are a little different on Windows in a way that I don't remember (I think it's that terminate / kill end up being basically the same thing).  There was at one point a whole helper routine devoted to trying to figure out how to do the terminate/kill in a way that was friendly to Python 2/3 and to Windows and *nix, the "cancel" function in util/launcher.py.  I'm not married to that function, but the logic there is all there for a reason -- this is something to treat with a little care.

shuds13:
> it doesn't seem like we can tell the signal to send. Can we just define/use our own function

We can define our own function. If we do this we may want to inherit from multiprocessing, and then we can define a function (like kill) if its not already there. We could then just add (with a reference) the code from the routine kill or any other that doesn't make use of python version specific features.

shuds13:
> Recommended protocol for Unix-based system is to try to send a terminate request (SIGTERM), which can be caught by the application to do clean-up; and then send a kill signal (SIGKILL) if the application remains unresponsive after SIGTERM.

The issue with petsc is that it calls mpi_abort in the SIGTERM signal handler, taking out everything (if are using mpi4py). We could use SIGINT instead of SIGTERM.

PR #478, opened 2020-06-11
removing old code that still hangs in an existing example

jmlarson1:
Although we tried to "be nice" and give back points until the run finished, we are still getting hangs in some use cases. 

Rather, we should just terminate the run (as we are not even processing the information from it).


jmlarson1:
@shuds13 The use case that started this issue does not correctly with `develop`. (It hangs when trying to given up to 1000 function values at the end of the run.)

I thought we could just remove that part of `destroy`, but I see there are still some issues with MPI tests failing (or going really slow) on Travis. 

One example is ` test_persistent_aposmm_timeout.py`, which runs on `develop` but not on this branch.

shuds13:
From what I can see test_persistent_aposmm_dfols.py is the test taking the longest time - is that right?

jmlarson1:
That test is the slowest test, but not because of the `destroy` routine. 

It took 120 seconds in the most recent 3.5, MPI test: 
https://travis-ci.org/github/Libensemble/libensemble/jobs/697068983
on this branch, but it took 969 seconds on `develop`:
https://travis-ci.org/github/Libensemble/libensemble/jobs/697293298

(Not sure why it's *that* long on `develop`.) I just reduced it's sim_max for debugging here.

But can you confirm that on your local system (on this branch) that the following test:

` mpiexec -np 4 python3 test_persistent_aposmm_timeout.py`

fails to complete, but prints out (at the very end):

```
[1]PETSC ERROR: ------------------------------------------------------------------------
[1]PETSC ERROR: Caught signal number 15 Terminate: Some process (or the batch system) has told this process to end
[1]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[1]PETSC ERROR: or see https://www.mcs.anl.gov/petsc/documentation/faq.html#valgrind
[1]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple Mac OS X to find memory corruption errors
[1]PETSC ERROR: likely location of problem given in stack below
[1]PETSC ERROR: ---------------------  Stack Frames ------------------------------------
[1]PETSC ERROR: Note: The EXACT line numbers in the stack are not available,
[1]PETSC ERROR:       INSTEAD the line number of the start of the function
[1]PETSC ERROR:       is given.
application called MPI_Abort(MPI_COMM_WORLD, 59) - process 1


```

shuds13:
Yes I get the error on the timeout test, unless I include the lines:

    import libensemble.gen_funcs
    libensemble.gen_funcs.rc.aposmm_optimizers = 'nlopt'

in calling script. So looks like petsc is getting imported in the subprocess as well and when call terminate we are taking it out. Presumably if the old lines are there it brings you out without the terminate.

I will think about best solution. But one answer is we check if petsc is imported and if so we do the hardkill, or replace terminate with SIGINT. Or we set the hard kill when we import petsc.

This has bought up the point that we DO need an APOSMM test that does not have the rc lines! Or maybe we should make these lines required.

jmlarson1:
I understand now. Yes, PETSc is getting imported because there aren't the `rc` lines in the test. 

I think that this can be our test without the `rc` lines (we should probably have one). 

I'm not sure what's the "best practice" but it makes some sense to set the hardkill when importing any package that doesn't support terminate. But setting such a variable at the top of the file seems like an odd place to do so (in my mind). But I'll let you decide what is best to do.

shuds13:
I've restructured a bit so if petsc4py.PETSc is loaded it will use SIGINT rather than SIGTERM. I moved the send_signal functions to module level, but maybe should move this to utils or something. 

I still have the hard kill if the SIGINT/SIGTERM times out, but it probably wont get called in the tests.

jmlarson1:
I think that that looks reasonable. Let's see how the tests perform on Travis. 

The most-recent version of this branch correctly terminates Chris Hall's use case (that first brought this issue to our attention).

shuds13:
@jmlarson1 
How do you feel about having **psutil**  as another dependency for APOSMM?

This provides the kill method that seems to work most reliably, and the only method
I've found so far that works reliably for external_localopt:

    import psutil
    parent = psutil.Process(process.pid)
    for child in parent.children(recursive=True):
        child.kill()
    parent.kill()

This could be used in place of the hard kill (the current one does not work for external_localopt as it does not kill the subprocess).

That said, do we need to send SIGTERM/SIGINT at all. Do we need to give the opportunity of graceful shutdown, as we are not getting anything out? 

The simplest way might be to just use the above for everything and make psutil always a dependency.

_Note: I did try catching SIGTERM in external_localopt and shuting down subprocess there, but that only worked intermittently._

jmlarson1:
@shuds13 It seems that `test_persistent_aposmm_external_localopt` is now the issue, but the `destroy` routine finishes without issue. (It hangs for me locally, as well as on Travis.)

I've tried sending `SIGINT` as we do for PETSc but that doesn't correctly kill the external process either.

jmlarson1:
> How do you feel about having **psutil** as another dependency for APOSMM?

That seems just fine to me. Feel free to push the simplified destroy.

shuds13:
Do you want to keep the terminate attempt or just always use the kill?

jmlarson1:
I would think that just killing makes the most sense, unless there is some risk in being that harsh. We don't want any more output from the subprocess. As each subprocess seems to handle terminate requests in different ways, a simple kill is what I'd like.

shuds13:
Do we still need the check in destroy:

    if not isinstance(previous_x, ConvergedMsg

Can we just kill the subprocess if its still active:

    if self.process.is_alive():

Then we don't need to pass the points. And on an APOSMM exception we can just loop over processes calling destroy.

jmlarson1:
Yes, I think that `.is_alive` is enough to check.

libensemble/gen_funcs/persistent_aposmm.py
```diff
@@ -667,14 +673,15 @@ def add_k_sample_points_to_local_H(k, user_specs, persis_info, n, comm, local_H,
     return persis_info
 
 
-def clean_up_and_stop(local_H, local_opters, run_order):
+# def clean_up_and_stop(local_H, local_opters):
+def clean_up_and_stop(local_opters):
```
shuds13, line 677: @jmlarson1 do you still want the option to print local_H information here?

jmlarson1: I'm fine with removing that (as I can print where `clean_up_and_stop` is called).

shuds13:
Coverage in node_resources.py has decreased as **psutil** is installed now in travis. I had psutil as an optional use in that module to avoid the dependency in libensemble main. Maybe we should reconsider this and whether to make psutil a required libensemble dependency.

jmlarson1:
I am fine with adding **psutil**. It seems like a useful utility when available. 

https://pypistats.org/packages/psutil

shuds13:
> I am fine with adding **psutil**. It seems like a useful utility when available.

Ok, I suggest we do this as a separate pull request and should warn users in develop they will need this dependency. Then we can remove the unnecessary lines in node_resources.

To my mind this one is ready.

libensemble/gen_funcs/persistent_aposmm.py
shuds13: Ok, be aware, I have moved it to a finally clause, so it is called on exit, including when there are exceptions.

PR #495, opened 2020-07-28
Feature/register apps

shuds13:
Add registration of applications by name to Executor.

### DIFF
--- a/docs/executor/overview.rst
+++ b/docs/executor/overview.rst
@@ -96,14 +96,15 @@ equivalent lines in the above example would be:
 
 Calling script::
 
-    exctr.register_calc(full_path='/path/to/my/exe', name='forces_app')
+    exctr.register_calc(full_path='/path/to/my/exe', app_name='forces_app')
 
 User sim func::
 
     task = exctr.submit(app_name='forces_app', num_procs=8, app_args='input.txt',
                         stdout='out.txt', stderr='err.txt')
 
-The name can be any identfier. This allows multiple applications to be registered.
+The ``app_name`` can be any identfier, while ``full_path`` is the application to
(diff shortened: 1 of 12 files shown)


shuds13:
> Looks good. Executor overview (`docs/executor/overview.rst`) may need some small updates.

Good point. I am updating.

shuds13:
This also makes me think about argument naming here. For I have used 'name' for register_calc, and 'app_name' in submit. @jlnav @jmlarson1 Any preferences here?

Calling script::

    exctr.register_calc(full_path='/path/to/my/exe', name='forces_app')

User sim func::

    task = exctr.submit(app_name='forces_app', num_procs=8, app_args='input.txt',
                        stdout='out.txt', stderr='err.txt')

Not to mention the function should really be register_app.

jlnav:
> This also makes me think about argument naming here. For I have used 'name' for register_calc, and 'app_name' in submit. @jlnav @jmlarson1 Any preferences here?

I think both parameters could be ``app_name``, but if this correction is made now, it should be done in such a way so that it isn't a breaking change.

I also support renaming ``register_calc`` to ``register_app``, but maybe this should be saved for before a major release.

shuds13:
The name/app_name is a new arg, so not a breaking change either way. So just which one is preferred. I think they should be the same. The register_calc/register_app would be a breaking change, so I won't be making that now.

jlnav:
> The name/app_name is a new arg, so not a breaking change either way. So just which one is preferred. I think they should be the same. The register_calc/register_app would be a breaking change, so I won't be making that now.

Ah, right. I like ``app_name``, it's more descriptive without being much longer.

Issue #502, opened 2020-08-08
multiprocessing error on macOS with Python3.8


cchall:
I recently updated to Python 3.8 on my Mac and found that libEnsemble examples no longer seemed to run when multiprocessing is being used. Trying examples/tutorials/tutorial_aposmm.py with `python tutorial_aposmm.py  --nworkers 3 --comms local` gives the following error:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/multiprocessing/spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/multiprocessing/spawn.py", line 125, in _main
    prepare(preparation_data)
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/multiprocessing/spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/multiprocessing/spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/runpy.py", line 262, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/runpy.py", line 95, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/Users/chall/.pyenv/versions/3.8.0/lib/python3.8/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/chall/research/github/libensemble/examples/tutorials/tutorial_aposmm.py", line 45, in <module>
    H, persis_info, flag = libE(sim_specs, gen_specs, exit_criteria, persis_info,
  File "/Users/chall/.pyenv/versions/venv380/lib/python3.8/site-packages/libensemble/libE.py", line 138, in libE
    return libE_funcs[comms_type](sim_specs, gen_specs, exit_criteria,
  File "/Users/chall/.pyenv/versions/venv380/lib/python3.8/site-packages/libensemble/libE.py", line 313, in libE_local
    wcomms = start_proc_team(nworkers, sim_specs, gen_specs, libE_specs)
[...]

PR #503, opened 2020-08-11
Enforce multiprocessing start method as 'fork' on macOS

jlnav:
Addresses #502 

This fix should address the problem in the above issue.

As stated on the issue, Python 3.8's multiprocessing module on macOS sets the new default launch method to 'spawn' instead of 'fork'. They made this change due to potential crashes on newer versions of macOS with 'fork'. However, I haven't observed any such crashes locally or on Travis, nor have users reported any (yet). Furthermore, in my limited testing 'spawn' produced semaphore leaking errors or RuntimeErrors, and it's my understanding that properly supporting 'spawn' would require mac users to refactor their calling scripts.

Please feel free to provide feedback or additional information about this issue.

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -309,6 +309,15 @@ def libE_local(sim_specs, gen_specs, exit_criteria,
 
     hist = History(alloc_specs, sim_specs, gen_specs, exit_criteria, H0)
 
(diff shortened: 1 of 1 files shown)


shuds13:
Looks good to me. I guess travis is still limited to testing on one version of macOS.

PR #504, opened 2020-08-13
Conda-install PETSc / petsc4py, add OSX Python 3.8 jobs

jlnav:
Conda-installing petsc4py seems to work on Python 3.8. OSX Python 3.8 jobs added (and allowed to fail).


jlnav:
It appears many of the dependencies with Python 3.8 on OSX still have issues. However, those jobs are in the Allowed Failures section. In any case, this PR confirms Python 3.8 compatibility on Linux.

PR #506, opened 2020-08-19
Feature/zrw docs logs

jlnav:
- Removes several older diagrams and their .xml source files
- Moves remaining .xml source files into separate directory
- Adds zero-resource worker description with diagrams to docs
- local_nodelist DEBUG messages slightly more descriptive (helps with testing)


libensemble/resources/resources.py
```diff
@@ -356,9 +356,10 @@ def get_local_nodelist(num_workers, workerID, resources):
 
         if workerID in zero_resource_list:
             local_nodelist = []
+            logger.debug("Worker {} is a zero-resource worker")
         else:
             index = WorkerResources.map_workerid_to_index(num_workers, workerID, zero_resource_list)
             local_nodelist = split_list[index]
+            logger.debug("Worker {}'s local_nodelist is {}".format(workerID, local_nodelist))
```
shuds13, line 363: I'm fine with this if it makes it clearer, but note that worker ID should appear at the start of the line.

docs/platforms/platforms_index.rst
```diff
@@ -134,6 +134,36 @@ available nodes where possible. ``jsrun`` can also queue runs. However, on
...
+configurations that associated persistent workers are still automatically assigned
+system resources. This can be wasteful if those workers only run in-place and
+don't use the Executor to submit applications to allocated nodes:
+
+.. image:: ../images/persis_wasted_node.png
+    :alt: persis_wasted_node
+    :scale: 40
+    :align: center
+
+This can be resolved within the Executor definition in a calling script. Set the
```
shuds13, line 150: in _the_ calling script.

docs/platforms/platforms_index.rst
```diff
@@ -134,6 +134,36 @@ available nodes where possible. ``jsrun`` can also queue runs. However, on
 other cluster and multi-node systems, if auto-resources is disabled, then runs without
 a hostlist or machinefile supplied may be undesirably scheduled to the same nodes.
 
+Zero-resource workers
+~~~~~~~~~~~~~~~~~~~~~
+
+Users with persistent ``gen_f`` functions may notice with certain
+configurations that associated persistent workers are still automatically assigned
```
shuds13, line 141: Wondering if can simplify the sentence. I'm wondering are there configurations where this is not the case? Could it be just: 

Users with persistent ``gen_f`` functions may notice that the persistent workers are still automatically assigned....

docs/platforms/platforms_index.rst
shuds13: This made me think - is calling script in the glossary? I feel like a user could confuse it with the batch script?

libensemble/resources/resources.py
jlnav: Ah, you're right. I accidentally overlooked those IDs. Thanks

docs/platforms/platforms_index.rst
jlnav: Good point, calling script currently isn't in the glossary, and should be added and clarified. 

I was also thinking users incorporating libensemble as part of other modules, routines, or studies might prefer not to think of their code as a "libensemble calling script", like how we don't consider our code numpy or psutil calling scripts. But this is just unimportant nitpicking!

PR #510, opened 2020-09-03
Docs/function guide

### LINKED ISSUE #485: "Best Practices" or Writers Guide for User Functions
In documentation hierarchy: User Guide -> Programming with libEnsemble -> (new) User Function Writers Guide

Details:
- flake8 or other standards for inclusion in libEnsemble (if requested)
- Standards for documenting user function
- Describe separation of calculation and parsing of libE data structures (e.g. six_hump_camel)
- etc.

[more ideas please]

jlnav:
Closes #485 

- WIP User function guide
- Various tags in docs
- Autodocs for gen_support & alloc_support functions
- Tentatively remove unused gen_support function parameter (?)


jlnav:
First draft completed.

View here: https://libensemble.readthedocs.io/en/docs-function_guide/function_guide.html

docs/function_guide.rst
```diff
@@ -0,0 +1,264 @@
+==================================
+Writing libEnsemble User Functions
+==================================
+
+libEnsemble coordinates ensembles of calculations performed by three main
+functions: a :ref:`Generator Function<api_gen_f>`, a :ref:`Simulator Function<api_sim_f>`,
+and an :ref:`Allocation Functions<api_alloc_f>`, or ``gen_f``, ``sim_f``, and
+``alloc_f`` respectively. These are all referred to as User Functions. Although
+libEnsemble includes several ready-to-use User Functions like
+:doc:`APOSMM<examples/aposmm>`, it's expected that most users will write their own.
```
shuds13, line 10: most users -> many users?
I think many are likely to reuse allocation functions, not sure what proportion of users will re-use gens.

docs/function_guide.rst
```diff
@@ -0,0 +1,264 @@
...
+most users, those who want to fine-tune how data is passed to their ``gen_f``
+and ``sim_f`` can write their own. The ``alloc_f`` is unique since it is called
+by the libEnsemble's manager instead of a worker.
+
+Most ``alloc_f`` function definitions written by users resemble::
+
+    def my_allocator(W, H, sim_specs, gen_specs, alloc_specs, persis_info):
+
+Where :doc:`W<data_structures/worker_array>` is an array containing information
+about each worker's state, and ``H`` is the *entire* History array available to
```
shuds13, line 165: The H sent to the allocation function is the trimmed H, it only contains rows intialized by the generator. So I guess its pretty much the entire useful array.

docs/function_guide.rst
```diff
@@ -0,0 +1,264 @@
...
+.. autofunction:: test_any_gen
+
+.. currentmodule:: libensemble.tools.alloc_support
+.. autofunction:: count_gens
+
+.. currentmodule:: libensemble.tools.alloc_support
+.. autofunction:: count_persis_gens
+
+Descriptions of included allocation functions can be found :doc:`here<examples/alloc_funcs>`.
+Below is the ``fast_alloc`` allocation function as an example:
```
shuds13, line 261: Maybe worth relating this to the default alloc_function - where the default (give_sim_work_first) has similar structure as fast_alloc but has some extra features that can be used in advanced cases.

docs/function_guide.rst
```diff
@@ -0,0 +1,264 @@
...
+writing different kinds of User Functions, and common development patterns.
+
+Generator Functions
+===================
+
+As described in the :ref:`API<api_gen_f>`, the ``gen_f`` is called by a
+libEnsemble worker via the following::
+
+    out = gen_f(H[gen_specs['in']][sim_ids_from_allocf], persis_info, gen_specs, libE_info)
+
```
shuds13, line 21: Somewhere here might mention the case of persistent gens, where we currently tend to have gen_specs empty and let the allocation function determine what is sent (though we might want to change this).

docs/function_guide.rst
```diff
@@ -0,0 +1,264 @@
...
+    ... # Perform simulation calculations
+
+    return local_H_out, persis_info
+
+Descriptions of included simulator functions can be found :ref:`here<examples/sim_funcs>`.
+
+The :ref:`Simple Sine tutorial<tutorials/local_sine_tutorial>` is an
+excellent introduction for writing simple user functions and using them
+with libEnsemble.
+
```
shuds13, line 141: I would mention the use of calc_status in sim functions.

docs/function_guide.rst
jmlarson1: I would say something like "we expect most users will to be able to adjust existing user functions for their use cases."

docs/function_guide.rst
jmlarson1: Agree on mentioning the difference between the default gens (which only have whatever was given to/from H to know about the past gen/sim work) and persistent gens (which occupy a worker for a period of time and do not need to be re-initialized every time their past work is completed and returned from the manager).

docs/function_guide.rst
jmlarson1: Definitely worth relating to the default alloc_f. Also note that if the alloc_f returns a 1 in the third argument, it lets the run stop. Otherwise, the alloc_f returning nothing is an error (when all workers are idle).

jlnav:
Suggestions have been addressed

docs/function_guide.rst
```diff
@@ -0,0 +1,299 @@
+==================================
+Writing libEnsemble User Functions
+==================================
+
+libEnsemble coordinates ensembles of calculations performed by three main
+functions: a :ref:`Generator Function<api_gen_f>`, a :ref:`Simulator Function<api_sim_f>`,
+and an :ref:`Allocation Functions<api_alloc_f>`, or ``gen_f``, ``sim_f``, and
+``alloc_f`` respectively. These are all referred to as User Functions. Although
```
shuds13, line 8: Allocation function (no s)

docs/function_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+dictionary containing pre-defined parameters for the ``gen_f``, and ``libE_info``
+is a dictionary containing libEnsemble-specific entries. See the API above for
+more detailed descriptions of the parameters.
+
+.. note::
+
+    If the ``gen_f`` is a persistent generator, then ``gen_specs`` will often be
+    empty if the ``alloc_f`` determines what fields to send to the generator.
+
+Typically users start by parsing their custom parameters initially defined
```
shuds13, line 40: I would probably say extract rather than parse - just my preference here.

docs/function_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+within ``gen_specs['user']`` in the calling script and defining a *local* History
+array based on the datatype in ``gen_specs['out']``, to be returned. For example::
+
+        batch_size = gen_specs['user']['batch_size']
+        local_H_out = np.zeros(batch_size, dtype=gen_specs['out'])
+
```
shuds13, line 46: It might be a bit advanced here. But it is possible, if you want to update existing fields to provide a sim_id and some subfield and it will just update that field. This is how I cancelled simulations from the gen in cwp that have already been generated.

H_o = np.zeros(len(sim_ids_to_cancel), dtype=[('sim_id', int), ('cancel', bool)])

@jmlarson1 any opinion on if we should include here.

shuds13: Perhaps we should wait for a future version. As this also goes with a more irregular comms pattern.

docs/function_guide.rst
```diff
@@ -0,0 +1,299 @@
...
+``alloc_f``, :doc:`persis_info<data_structures/persis_info>` is a dictionary
+containing state information, :doc:`gen_specs<data_structures/gen_specs>` is a
+dictionary containing pre-defined parameters for the ``gen_f``, and ``libE_info``
+is a dictionary containing libEnsemble-specific entries. See the API above for
+more detailed descriptions of the parameters.
+
+.. note::
+
+    If the ``gen_f`` is a persistent generator, then ``gen_specs`` will often be
+    empty if the ``alloc_f`` determines what fields to send to the generator.
```
shuds13, line 38: gen_specs['in']

shuds13:
I think this looks pretty good. Maybe consider making sub-pages for gen/sim/alloc? as its a bit long in one page.

jlnav:
Additional suggestions have been addressed, including splitting up guide into parts. I also moved it into the "Programming with libEnsemble" section in the docs, under the user function API:

https://libensemble.readthedocs.io/en/docs-function_guide/function_guides/function_guide_index.html

shuds13 (approved):

jmlarson1 (approved):

PR #512, opened 2020-09-24
Handling of large history arrays

jmlarson1:
Numpy now returns a view for the entire history array when indexing multiple fields. From the numpy [documentation:](https://numpy.org/doc/stable/user/basics.rec.html#accessing-multiple-fields)

>  Note that unlike for single-field indexing, the dtype of the view has the same itemsize as the original array, and has fields at the same offsets as in the original array, and unindexed fields are merely missing.
 
So an object the size of the entire history is communicate every time we send multiple fields. This can be slow (or prohibitive) when the history is large, we could send a sequence of single-fields out of the history. Or we can apply the [repack_fields](https://numpy.org/doc/stable/user/basics.rec.html#numpy.lib.recfunctions.repack_fields) option.

- A downside of this option is that it creates a structure containing copy of the fields to be sent. 
- A downside of a sequence of single-field sends is the communication cost. 

I vote for the copy approach, which is what will be in this pull request.

### DIFF
--- a/examples/misc/ps_nodes.sh
+++ b/examples/misc/ps_nodes.sh
@@ -27,4 +27,3 @@ do
     ssh $hname "hostname; ps aux|grep $uname|grep '$appname'"
   fi
 done
-
--- a/libensemble/libE_manager.py
+++ b/libensemble/libE_manager.py
@@ -9,6 +9,7 @@ import glob
 import logging
 import socket
 import numpy as np
+import copy
 
 from libensemble.utils.timer import Timer
 from libensemble.message_numbers import \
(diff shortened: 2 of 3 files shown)


jlnav (approved):
Approved. Maybe I misunderstand, but communicating each field individually would _also_ produce new-structure overhead besides the communication costs, right?

jmlarson1:
I'm not sure, but I thought that we could communicate `H['x']` without having the view of the entire `H`. The problem before was we were communicating `H[['x']]` and that inside list meant we got a full view of `H`. But I assume the local copy of `H` on the manager will be faster than the multiple communications.

shuds13:
I agree that using repack_fields is the better solution which should then do the same as numpy 1.15 behavour. Note that it looks like this function was only introduced in numpy 1.15, so earlier versions of numpy will not work (unless we do a version test in the code and do it conditionally).

jmlarson1:
The last version of numpy 1.14 was released in 2018 (not that long ago). We should either do a version test in the manager for the communication or assert that numpy is >= 1.15. What do you think?

shuds13:
I think a version check is okay. I think other than this we support quite old versions of numpy. At least in libensemble core modules.

jmlarson1:
Great. I have attempted such an numpy version check and then check if `repack_fields` is in `dir()` before using it. Open to suggestions if this isn't the proper way to do such a check.

shuds13:
Looks fine to me.

shuds13 (approved):

PR #514, opened 2020-10-02
Update MANIFEST.in and relevant docs

shuds13:
Updates to MANIFEST.in and and changes to release process

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,8 +4,10 @@ include *.rst
 include LICENSE
 include CONTRIBUTING
 include install/*
+include .flake8
 
 recursive-include libensemble *
+recursive-include postproc_scripts *
 recursive-include examples *
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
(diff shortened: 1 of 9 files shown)


shuds13:
@jmlarson1 @jlnav What do you think of the `+dev` proposal. I want to be able to identify by version whether code is a release or in development. 

There is a slight problem with this suggestion, that if you take a release branch from develop, and then some changes are made on develop (for a future release) before release branch is merged back into develop, then those changes will have the previous version number (with +dev) even though changes are afterwards. 

In any case, I think its better than what we have now where \__version__ does not distinguish a release from develop.

jmlarson1:
I'm fine with such a usage. Is this useful only in the release process, or for all branches?

shuds13:
I was thinking only to distinguish releases from "in development", so the dev is not just develop branch, but would be any non-release code. I don't like that an installation would say v0.7.0 if it is not the release. I'm not sure it would be good to use at more granularity.

jmlarson1:
Thanks. I think it is good that the development version number is, for example v0.7.1_dev (or whatever) as opposed to being v0.7.0 (which differs a bit from what was in the v0.7.0 release many months prior)

shuds13:
The reason for `+dev` is the PEP standard for local builds https://www.python.org/dev/peps/pep-0440/#local-version-identifiers

I think this is closest to how we want to use it and looks like some other projects use it that way.

PR #515, opened 2020-10-06
Trying to add a flag so regression tests can be run with warnings

jmlarson1:
Appending to `$PYTHON_FLAGS` variable may not be the best way to start passing flags to python.

If there is a more robust or scalable way to do this, feel free to adjust or harmonize the various arguments in this script. (It seems to be getting a bit large, but perhaps that's necessary for a unified testing script with many options.)

### DIFF
--- a/libensemble/tests/run-tests.sh
+++ b/libensemble/tests/run-tests.sh
@@ -203,7 +203,7 @@ usage() {
   echo "  -t              Run the regression tests using TCP comms"
   echo "  -p {version}    Select a version of python. E.g. -p 2 will run with the python2 exe"
   echo "                  Note: This will literally run the python2/python3 exe. Default runs python"
-  echo "  -W {name}       Supply warnings level for tests"
+  echo "  -A {-flag arg}  Supply arguments to python"
   echo "  -n {name}       Supply a name to this test run"
   echo "  -a {args}       Supply a string of args to add to mpiexec line"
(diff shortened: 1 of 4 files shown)


libensemble/tests/run-tests.sh
```diff
@@ -261,6 +263,10 @@ while getopts ":p:n:a:y:hcszurmlt" opt; do
       echo "Running with user supplied test list"
       export REG_TEST_LIST="$OPTARG"
       ;;
+    W)
+      echo "Warning level parameter: $OPTARG" >&2
+      PYTHON_FLAGS='-W error'
```
jlnav, line 268: Should this be:
``PYTHON_FLAGS="-W $OPTARG"`` ?

Otherwise it doesn't matter what the required argument is, it will be ``error``.

jmlarson1: Yes, I think you are right.

shuds13:
I think in run_tests I would do:
PYTHON_RUN="python$PYTHON_VER $PYTHON_FLAGS"
rather than add on every line.

Also, we could make it more general python args you could make it -A and then do eg. 
./run-tests.sh -A "-W error"

and
PYTHON_FLAGS="$OPTARG"

shuds13:
or PYTHON_FLAGS="$PYTHON_FLAGS $OPTARG"

jmlarson1:
@shuds13 I like that more. Does the new version seem right?

shuds13 (approved):
Looks good.

PR #517, opened 2020-10-13
Trying out multi-obj with DEAP

jmlarson1:


jmlarson1:
Unless @nneveu thinks something needs to be changed, this seems to successfully adjust the `deap` regression test to perform multi-objective optimization.

PR #522, opened 2020-10-16
Feature/async uniform sampling

jmlarson1:


shuds13 (approved):

shuds13:
@jmlarson1 

Might be worth comparing to changes I made in this branch 

https://github.com/Libensemble/libensemble/blob/cwp/async_build/libensemble/alloc_funcs/start_only_persistent.py

as I realise this is the same allocation function. That version has a number of other changes, but there is slightly different terminology used.

PR #534, opened 2020-10-21
Fix crash when running on Windows

AngelFP:
Hi all,

I have been testing a bit libEnsemble for a project here at DESY and so far I'm very impressed with it. I have however found a small issue when trying to run it also on my Windows laptop.

This PR fixes a small bug that causes libEnsemble to crash when running on a Windows system.

The crash happens because of a call to `os.uname` (introduced in https://github.com/Libensemble/libensemble/pull/503), which is only available in Unix systems. The fix here simply changes this call to `platform.system`, which works also on Windows. I have not been able to test this myself, but in principle `platform.system()` should still return `'Darwin'` when running on a Mac.

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -16,6 +16,7 @@ is dumped to file, and MPI abort is called.
 __all__ = ['libE']
 
 import os
+import platform
 import logging
 import random
 import socket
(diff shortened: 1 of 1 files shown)


jmlarson1:
Hi @AngelFP We're glad libEnsemble is helping you and thank you for the PR. Feel free to reach out with any other questions you may have. We are happy to consider user needs when adding features.

PR #537, opened 2020-10-27
Feature/better h0 support

jmlarson1:
Before `H0` was assumed to contain only points with `given` and `returned` as `True`. 

A way to do simulation evaluations at some pre-existing set of points is to put them into `H0` and have an `alloc_f` give them out (without needed a `gen_f` call). 

In this regression test, libEnsemble was marking them as `given` and `returned` but the `alloc_f` was giving them out anyway (so the test was performing correctly).

### DIFF
--- a/libensemble/alloc_funcs/give_pregenerated_work.py
+++ b/libensemble/alloc_funcs/give_pregenerated_work.py
@@ -11,8 +11,8 @@ def give_pregenerated_sim_work(W, H, sim_specs, gen_specs, alloc_specs, persis_i
     """
 
     Work = {}
-    if not persis_info:
-        persis_info['next_to_give'] = 0
+    # Unless already defined, initialize next_to_give to be the first point in H
+    persis_info['next_to_give'] = persis_info.get('next_to_give', 0)
 
     if persis_info['next_to_give'] >= len(H):
         return Work, persis_info, 1
--- a/libensemble/alloc_funcs/give_pregenerated_work.py
+++ b/libensemble/alloc_funcs/give_pregenerated_work.py
@@ -14,11 +14,15 @@ def give_pregenerated_sim_work(W, H, sim_specs, gen_specs, alloc_specs, persis_i
     if not persis_info:
         persis_info['next_to_give'] = 0
 
-    assert persis_info['next_to_give'] < len(H), 'No more work to give inside give_pregenerated_sim_work.'
+    if persis_info['next_to_give'] >= len(H):
+        return Work, persis_info, 1
 
     for i in avail_worker_ids(W):
         # Give sim work
         sim_work(Work, i, sim_specs['in'], [persis_info['next_to_give']], [])
         persis_info['next_to_give'] += 1
+
+        if persis_info['next_to_give'] >= len(H):
+            break
 
     return Work, persis_info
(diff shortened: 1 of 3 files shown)


libensemble/history.py
```diff
@@ -54,17 +54,25 @@ def __init__(self, alloc_specs, sim_specs, gen_specs, exit_criteria, H0):
...
-        # Prepend H with H0
-        H['sim_id'][:len(H0)] = np.arange(0, len(H0))
-        H['given'][:len(H0)] = 1
-        H['returned'][:len(H0)] = 1
+            if 'given' not in fields:
+                print("Marking entries in H0 as having been given and returned", flush=True)
+                H['given'][:len(H0)] = 1
+                H['returned'][:len(H0)] = 1
+            elif 'returned' not in fields:
+                print("Marking entries in H0 as having been returned", flush=True)
```
shuds13, line 70: I would make the print statements logger.manager_warning - thoes get echoed to std err by default. But we are bound by xsdk to not have print in core libe.

libensemble/history.py
```diff
@@ -54,17 +54,25 @@ def __init__(self, alloc_specs, sim_specs, gen_specs, exit_criteria, H0):
...
+                H['returned'][:len(H0)] = 1
```
shuds13, line 71: What if they set given to False? Perhaps returned should equal given?

libensemble/alloc_funcs/give_pregenerated_work.py
```diff
@@ -14,11 +14,15 @@ def give_pregenerated_sim_work(W, H, sim_specs, gen_specs, alloc_specs, persis_i
     if not persis_info:
         persis_info['next_to_give'] = 0
 
-    assert persis_info['next_to_give'] < len(H), 'No more work to give inside give_pregenerated_sim_work.'
+    if persis_info['next_to_give'] >= len(H):
+        return Work, persis_info, 1
 
     for i in avail_worker_ids(W):
         # Give sim work
         sim_work(Work, i, sim_specs['in'], [persis_info['next_to_give']], [])
```
shuds13, line 22: Should it check 'given' for the row its on - and increment if True?

libensemble/history.py
jmlarson1: I agree and have implemented this.

libensemble/alloc_funcs/give_pregenerated_work.py
jmlarson1: Yes. That is possible. This allocation function is called `give_pregenerated_sample` so I think that it's okay to just go through all of the rows of `H` in order. (But I'm fine with checking `given`.)

Note that the top of this allocation function is not very robust: If there is `persis_info` then `next_to_give` will not be initialized. I hope that such issues will be addressed in our re-working of allocation functions.

libensemble/history.py
jmlarson1: Sounds good. This has been done in 102461c8

libensemble/alloc_funcs/give_pregenerated_work.py
shuds13: Should it just be checking the 'next_to_give' component is initialized? Something like:

    persis_info['next_to_give'] = persis_info.get('next_to_give', 0) + 1

jmlarson1: The initialization is fixed in 8080b1b4

shuds13 (approved):

libensemble/alloc_funcs/give_pregenerated_work.py
jmlarson1: Thank you for the feedback!

PR #539, opened 2020-10-29
Experimental/protecting libE fields

jmlarson1:
This proposed change checks that any field returned from a `sim_f` or `gen_f` is not protected. (We could do a check in `check_inputs` that `gen_specs['out']` and `sim_specs['out']` aren't protected.)

Also, the default is to deepcopy the protected libE fields and ensure they aren't changed by the `alloc_f`. (This can be overwritten by setting `libE_specs['safe_mode'] = False`)

### DIFF
--- a/libensemble/history.py
+++ b/libensemble/history.py
@@ -96,7 +96,7 @@ class History:
 
         for j, ind in enumerate(new_inds):
             for field in returned_H.dtype.names:
-                assert field not in protected_libE_fields, "The field " + field + " is protected"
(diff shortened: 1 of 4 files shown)


shuds13 (approved):
Presumably last line should sat `libE_specs['safe_mode'] = False`

When `cancel_requested` is added we can make an exception.

Is there a simple way to unit test this?

jlnav (approved):

jmlarson1:
> Is there a simple way to unit test this?

I've added a unit test and removed `sim_id` from the protected field names as that can be given by example `gen_f`

PR #540, opened 2020-10-29
Testing/Resolve Travis dependency / testing issues with Python 3.8 on osx

### LINKED ISSUE #513: Investigate dependency issues for Python 3.8 on macOS Travis jobs
See: https://travis-ci.org/github/Libensemble/libensemble/builds/731087713

jlnav:
Small change on Travis so Python version is detected correctly for osx Python 3.8 jobs. Enforces multiprocessing start method for osx on necessary jobs and routines.

Perhaps the multiprocessing start method setting should be reduced to a helper function, instead of the same import and platform check occuring multiple times

With these changes, a recent build had all jobs pass through Python 3.8, Linux and osx:

https://travis-ci.org/github/Libensemble/libensemble/builds/739654275

Closes #513


jlnav:
The builds have passed, including all Allowed Failures, but I'm unsure why coverage dropped. No lines listed at Coverage are marked as "uncov", which is typically how I tell I've written new code not evaluated by a test. For instance, ``aposmm_localopt_support.py`` is marked as having decreased coverage, even though the single line (19) I added is covered:

https://coveralls.io/builds/34606765/source?filename=libensemble/gen_funcs/aposmm_localopt_support.py

PR #541, opened 2020-10-30
Updating libE_specs for safe_mode and other recent changes

jmlarson1:
Continuation of #539


jmlarson1:
@jlnav Thoughts on ordering the entries in the docstring in `libensemble/docs/data_structures/libE_specs.rst` to be easier to read? Perhaps matching the grouping or sorting in `libensemble/libensemble/tools/fields_keys.py`? 

(Just looking, we should consider not having `comms` and `comm` options in `libE_specs`)

docs/data_structures/history_array.rst
```diff
@@ -11,7 +11,9 @@ Fields in ``H`` include those specified in ``sim_specs['out']``,
 ``gen_specs['out']``, and ``alloc_specs['out']``. All values are initiated to
 0 for integers, 0.0 for floats, and False for Booleans.
 
-Below are the protected fields used in ``H``:
+Below are the protected fields used in ``H``. Other than ``'sim_id'``, these
+fields cannot be overwritten by user functions (unless
+``libE_spces['safe_mode']`` is set to ``True``).
```
jlnav, line 16: small typo

jlnav:
> @jlnav Thoughts on ordering the entries in the docstring in `libensemble/docs/data_structures/libE_specs.rst` to be easier to read? Perhaps matching the grouping or sorting in `libensemble/libensemble/tools/fields_keys.py`?

I can look into that. I've been thinking it would be nice if ``libE_specs`` in the docs was _also_ autodoced the way many other portions are. This doesn't have to be done now.

> (Just looking, we should consider not having `comms` and `comm` options in `libE_specs`)

``comms`` is usually populated within ``libE_specs`` by ``parse_args()``, I don't know where else we'd specify to libE what communication style is being used.

As for ``comm``, I believe we're required by xSDK to offer the capability to use a custom MPI communicator, which currently can be specified there. @shuds13 can speak more to this.

Unless you're asking if we should rename these fields to be more descriptive?

docs/data_structures/history_array.rst
jmlarson1: Thanks

jmlarson1:
> ...it would be nice if `libE_specs` in the docs was _also_ autodoced the way many other portions are. This doesn't have to be done now.

Can you give an example to show what the other portions of code that you are thinking of? (It seems autodoced to me.)

> Unless you're asking if we should rename these fields to be more descriptive?

That is one way to avoid confusion. `comm_type` and `mpi_comm` as opposed to `comms` and `comm`?

jlnav:
> Can you give an example to show what the other portions of code that you are thinking of? (It seems autodoced to me.)

Oops, I was mistaken thinking that any data structures are already autodoced

> That is one way to avoid confusion. `comm_type` and `mpi_comm` as opposed to `comms` and `comm`?

Sure! This would, however, be an impactful API change

shuds13:
I think there are some autodoc options in sphinx for variables - through comments. I tried them out at some point, but I'm not sure I liked the results. But that was a long time ago.

shuds13 (approved):

PR #547, opened 2020-11-02
Docs/disambiguating mpi comm

jmlarson1:
Trying to reserve the use of `comm` for a `comms` instance and use `mpi_comm` when referring to an mpi communicator.

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -203,8 +203,8 @@ def libE_mpi_defaults(libE_specs):
 
     from mpi4py import MPI
 
-    if 'comm' not in libE_specs:
-        libE_specs['comm'] = MPI.COMM_WORLD  # Will be duplicated immediately
+    if 'mpi_comm' not in libE_specs:
+        libE_specs['mpi_comm'] = MPI.COMM_WORLD  # Will be duplicated immediately
 
     return libE_specs, MPI.COMM_NULL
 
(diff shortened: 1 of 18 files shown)


libensemble/libE.py
```diff
@@ -270,8 +270,8 @@ def libE_mpi_worker(libE_comm, sim_specs, gen_specs, libE_specs):
     "Worker routine run at ranks > 0."
 
     from libensemble.comms.mpi import MainMPIComm
-    comm = MainMPIComm(libE_comm)
-    worker_main(comm, sim_specs, gen_specs, libE_specs, log_comm=True)
+    mpi_comm = MainMPIComm(libE_comm)
```
shuds13, line 273: This functions return is not mpi_comm. It is the libE comm object

shuds13: libE.py lines

    def __init__(self, comm):
        self.parent_comm = comm

could change the comm to mpi_comm:

    def __init__(self, mpi_comm):
        self.parent_comm = mpi_comm

shuds13: and maybe  comms_abort(comm):

shuds13 (commented):
Once my comments are addressed I think this is ready to pull in.

libensemble/libE.py
jmlarson1: All of these were addressed in 7ecae7e8

PR #552, opened 2020-11-05
[WIP] Collab/cwp calib tutorial

jlnav:


jlnav:
Current draft:

https://libensemble.readthedocs.io/en/collab-cwp_calib_tutorial/tutorials/cwp_calib_tutorial.html

shuds13:
@jlnav 
Can this be branched from cwp_standalone, so we can get rid of the old commits? Maybe it doesn't matter but I'm a bit concerned about getting in a tangle.

I will give some inline comments, but first here is my brain dump of info I would want to communicate at the start, some of which you probably already have. Ignore the bad grammer/wording. Moses/Stefan can help with improving this and getting terminology right.

Intro:
This tutorial is based on a CWP callibration example.....(*check Moses). The example test case, provided in libE, uses simplified standalone routines in place of the CWP library.

Outline of problem (Moses/Stefan can help improve this):
You have 'observed values' at a given set of points. In aiming to fit a model to these points, a parameterized function is evaluated at each point. One parameter-set is called a theta. The goal is to find the Theta that most closely matches observed values.
After a batch of randomly sampled values, a model (gaussian process) is used to generate new thetas. Each Theta is evaluated at each of the points, until the given threshold is reached. Based on evaluation results, we want the ability to cancel the evaluation previously requested Thetas, to improve efficiency.

Generator:
After an initial batch of thetas is explored, the generator uses evaluation results to construct the model. The model is used to create give out new Thetas. Note that each evaluation (row of H) is a theta at a given point - so (points x theta) evaluations are loaded into H).

Evaluation results are passed back to the generator as they complete. The generator can use these results to determine that previously issued thetas no longer need to be evaluated, and can request cancellation of the corresponding evaluations. If these evaluations are pending,  they will not get sent to the simulator. If they are currently running in a simulator, a kill signal will be sent. When these points have a result it may still get passed back to the generator, but the generator can choose to use or ignore the result.


This use-case uses the following features of libEnsemble (*maybe mention this further up)
 - (*active receive gen). There persistent generator remains in an active state until it becomes non-persistent - irregular send/receving of data (not ping-pong).
 - Generator cancellation of evaluations.

*provisional terminology (active receiver)

Other features we use, but maybe are not crucial points for this tutorial
 - We have option to build the model asynchronously in a background thread.
 - Prioritization of simulations.

jlnav:
@shuds13 Thank you for the excellent feedback,
> Can this be branched from cwp_standalone, so we can get rid of the old commits? Maybe it doesn't matter but I'm a bit concerned about getting in a tangle.

- [ ] Understood. I'll do that. This branch is originally off develop, so if develop is merged into cwp_standalone, that may clean up this PR a bit.

> Intro:
> This tutorial is based on a CWP callibration example.....(*check Moses). The example test case, provided in libE, uses simplified standalone routines in place of the CWP library.

- [ ] I still need to mention this detail.

> Outline of problem (Moses/Stefan can help improve this):
> You have 'observed values' at a given set of points. In aiming to fit a model to these points, a parameterized function is evaluated at each point. One parameter-set is called a theta. The goal is to find the Theta that most closely matches observed values.
> After a batch of randomly sampled values, a model (gaussian process) is used to generate new thetas. Each Theta is evaluated at each of the points, until the given threshold is reached. Based on evaluation results, we want the ability to cancel the evaluation previously requested Thetas, to improve efficiency.

- [ ]  More precise outline of mathematics in problem, describe observed values, gaussian process model fitting, parameterization.

- [ ] Describe justification for evaluation of previously requested Thetas

>     * Are there any nice graphics that will help show this?

- [ ]  Nice graphics to help show this

> Generator:
> After an initial batch of thetas is explored, the generator uses evaluation results to construct the model. The model is used to create give out new Thetas. Note that each evaluation (row of H) is a theta at a given point - so (points x theta) evaluations are loaded into H).
> 
> Evaluation results are passed back to the generator as they complete. The generator can use these results to determine that previously issued thetas no longer need to be evaluated, and can request cancellation of the corresponding evaluations. If these evaluations are pending, they will not get sent to the simulator. If they are currently running in a simulator, a kill signal will be sent. When these points have a result it may still get passed back to the generator, but the generator can choose to use or ignore the result.

- [ ]  Describe generator cancellation requests in the context if thetas no longer being needed in model construction

> This use-case uses the following features of libEnsemble (*maybe mention this further up)
> 
>  (*active receive gen). There persistent generator remains in an active state until it becomes non-persistent - irregular send/receving of data (not ping-pong).

- [ ] Active receive generator.

>* Generator cancellation of evaluations.
>

- [x] Generator cancellation of evaluations.

> 
> *provisional terminology (active receiver)
> 
> Other features we use, but maybe are not crucial points for this tutorial
> 
>     * We have option to build the model asynchronously in a background thread.
> 
>     * Prioritization of simulations.

- [ ] (Optional) asynchronously building of model

- [ ] (Optional) Simulation prioritization

PR #556, opened 2020-11-17
Starting to limit travis test when on develop

jmlarson1:


shuds13:
@jmlarson1 
you'll need to take tcp out of the matrix in .travis.yaml. Could just comment out for now.

PR #557, opened 2020-11-18
Testing/travis time experiments

jmlarson1:


libensemble/libE_manager.py
```diff
@@ -435,7 +435,10 @@ def _alloc_work(self, H, persis_info):
         fields before the alloc_f call and ensures they weren't modified
         """
         if self.safe_mode:
-            saveH = copy.deepcopy(H[protected_libE_fields])
+            if 'repack_fields' in globals():
+                saveH = copy.deepcopy(repack_fields(H[protected_libE_fields]))
```
shuds13, line 439: Do you need to do deepcopy and repack_fields in same line? Surely repack gives you deep copy.

jmlarson1:
@jlnav Do you know why the itemsize is still large when using repack_fields (for example) in the MemoryError on line 3858 of this Travis build? https://travis-ci.org/github/Libensemble/libensemble/jobs/744280837

jmlarson1:
I've added a unit test in 1c54230f to test that repack_fields does what we think it does.

jmlarson1:
@shuds13 The latest commit on this branch tracks the memory usage. For that regression test, there is a repeated point in the run where trying to get the itemsize allocates memory. See the output before the failure in this commit for an example.

https://api.travis-ci.org/v3/job/747695820/log.txt

I'm going to remove the `time=0.1` before merging as there are other parts of this branch that I'd like to get into develop

PR #559, opened 2020-11-20
Add generator function for discrete fidelities

AngelFP:
This PR adds a new generator function `persistent_gp_mf_disc_gen_f` which can handle multi-fidelity optimization with Dragonfly when the fidelity space is discrete. It also modifies the current `persistent_gp_mf_gen_f` function to allow the cost function to be specified by the user (i.e., it is no longer hard-coded) and to handle that the fidelity is now separate from the other input parameters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ code/tests/regression_tests/output/
 libensemble.egg-info
 docs/_build
 x.log
+.vscode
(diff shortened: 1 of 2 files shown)


jmlarson1:
Hi @AngelFP This looks great. Do you have a file that tests the capabilities of this generator function? (It can use a simple test sim_f instead of a full simulation.)

If you have a file, can you add it to `libensemble/libensemble/tests/regression_tests` so that it can be used in our testing suite (name it starting with `test_`).

shuds13:
I am going to merge this in, it is going to feature/multi_fidelity, so we can work out tests later.

AngelFP:
Hi @jmlarson1 and @shuds13, thanks for merging this already. I didn't have the time last days to check this PR, but I agree with the point about adding a test. This will certainly be done before we try to merge these changes into  the main `develop` branch.

PR #560, opened 2020-11-25
Improve SLURM nodelist detection

shuds13:
Ensure greater range of formats is supported.

Also need to determine exactly when shortnames are or are not required, so that this works on all SLURM systems.


shuds13:
Still need to make sure test for both shortname and long name cases.

Also cneed to heck where workers_per_node is calculated in mpi_resources. May be a bug there.

PR #561, opened 2020-12-01
Merge develop into multi_fidelity

shuds13:
@RemiLehe This will update with recent changes. This includes changes to reading the SLURM nodelist from the environment and a bug fix when using zero-resource workers.


RemiLehe:
Great! Thanks a lot for this update!

PR #563, opened 2020-12-03
Implement beginning GitHub actions

jlnav:
First implementation of GitHub's automation / CI alternative. This service is _**much**_ faster than Travis, recent runs can be evaluated by clicking the "Actions" tab in the repository.

Recent ~~successful~~ run: https://github.com/Libensemble/libensemble/actions/runs/399404521

A couple notes and thoughts:

- There's a github-actions annotation added to each run stating ``Warning: 'conda-forge' already in 'channels' list, moving to the top``, but removing conda-forge causes the installs to not work.
- Most of the conda installs have to happen sequentially, otherwise conda fails with ``Packagesnotfounderror``
- Balsam testing is not yet implemented, but _should_ be possible
- I wouldn't be opposed to re-adding TCP testing since these jobs complete so fast.
- ``test_fast_alloc.py`` passes consistently, maybe because these VMs have more RAM (7 GB)
- Starting a debug session is as easy as uncommenting the ``uses: mxschmitt/action-tmate@v3`` line in the ``.yml`` file, then pushing.

I still need to update the docs, add badges, and add Balsam testing, but maybe this can happen on another branch.


jmlarson1 (approved):
I really like this. It looks great and seems to run much faster. I think we should update the docs and badges in this PR. (I'd look to add other actions in a different PR.)

A few questions: 
 - Is there any harm in building with both travis and github actions?
 - What are the limitations on run-time and number of parallel process in github-actions? (Travis was 50min and 4)
 - Can we set up automated fail/success emails like Travis? 
 - Can we have github actions be used for PRs instead of travis
 - In this discussion https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/84 someone mentions the following to allow for single jobs to be re-run. 

> I think I found a workaround here. I have split my matrix build to multiple yaml files, each one with a different name but the same trigger. Each of them contain only a single run. As it looks, this enables me to re-run jobs individually. I do so by selecting “re-run all jobs”, where “all” is now always exactly one. The price you need to pay is that you have some code duplication and cannot use the matrix feature. For me, this is acceptable as the CI is actually done by a script and shared code…

jlnav:
>     * Is there any harm in building with both travis and github actions?

No harm from what I can tell, in the checks section both Actions and Travis are included together. Many projects build on multiple CI services.

>     * What are the limitations on run-time and number of parallel process in github-actions? (Travis was 50min and 4)

According to: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits

**6 hours limit and 20 concurrent jobs!?** But I might enforce a 1-hour limit for sanity. All the jobs (as of 12/4) completed within 30 minutes.

>     * Can we set up automated fail/success emails like Travis?

This *may* be a profile setting: https://github.com/settings/notifications . There's an section for Github Actions under Notifications.

>     * Can we have github actions be used for PRs instead of travis

The idea then would be: Actions for PRs and branches, and Travis only for branches, right?

>In this discussion https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/84 someone mentions the following to allow for single jobs to be re-run.

Seems like a good solution, but if we add Windows, Python 3.9, and TCP, then we'll need 36 files. Right now we would need 10. Maybe we start by splitting by python-version. I'll keep thinking about this.

shuds13 (approved):

PR #566, opened 2020-12-21
Experimental/ml keras example

jlnav:
Requires Tensorflow, Keras, and Horovod.

This test launches a Python Machine Learning training routine as a ``gen_f``
application through the Executor, then evaluates the resulting model in a ``sim_f``
using test data. The number of training epochs, MPI processes, arguments for the
Tensorflow application, and other parameters can be easily adjusted.

Presumably, any other simulation that incorporates a trained model could be used in the ``sim_f``. Another potential route to explore is have a ``gen_f`` generate sets of hyperparameters for the ML app, then launch the app in a ``sim_f``.

TODO:

- [ ]  Submission scripts


jlnav:
Closing until work/research can resume

PR #568, opened 2020-12-23
Docs/year and ver file

jmlarson1:


shuds13:
We still have the version in __init__.py, which I guess we can't use the VERSION file for where it is, as it would need to be in the package.

libensemble/__init__.py
```diff
@@ -4,7 +4,10 @@
 Library to coordinate the concurrent evaluation of dynamic ensembles of calculations.
 """
 
-__version__ = "0.7.1+dev"
+with open("VERSION", "r") as f:
```
jmlarson1, line 7: @shuds13 This is being invoked where the libE calling script is, not where `__init__.py` resides. We don't want to require a `VERSION` file where the calling script is. Is there a way to refer to the base directory of the package? Something like
`with open(basedir + "libensemble/VERSION", "r") as f:`
?

shuds13:
@jmlarson1 
I have made the '+dev' part of the version name in VERSION. If we want to have just X.Y in conf.py we could just split at the dots, but I havn't done that currently. I was thinking we could after v1.0, but not sure if that is used anywhere now.

So now the two places to change are in VERSION and in readme.txt. And I'm wondering if readme.txt, should just hold info for last release. 

Also, now VERSION file is in the libensemble dir, we don't need to add it to the MANIFEST, as it takes everything in that dir.

shuds13 (approved):

shuds13:
Okay, all tests failed, ,did I break something.

jmlarson1:
Yes you did :-)

I commented on the change in `__init__.py` to point out why everything is failing now.

libensemble/__init__.py
shuds13: Yes, just thinking what is the best method

shuds13:
@jmlarson1 
I think its actually simpler, and more standard, to use a version.py. What do you think?

shuds13:
and maybe we should squash this when merging in.

jmlarson1:
Sounds good. Looks good.

I tried to squash but I don't know if it worked. Feel free to pull in (and squash) when you like.

PR #569, opened 2021-01-05
Set initial capital of Dragonfly optimizer to match number of workers

AngelFP:
This PR fixes a crash which occurred whenever the number of workers was larger that the default size of the initialization sample of the Dragonfly optimizer (5 by default).

The size of the initial capital is now set to match the number of workers. This makes sure that we always have enough initial points for the first batch of runs. However, in the future, these two parameters (number of workers and of initial points) could also be set independently.


jmlarson1:
Looks great @AngelFP Thanks!

Issue #573, opened 2021-01-05
Remove 'libE_' prefixes from core modules


jlnav:
Since modules are undeniably components of libEnsemble and aren't typically user-facing and at risk of being confused for other modules.

jmlarson1:
This change has been made in #576. `logger` is possibly external facing... but I think the risk of confusion is near nothing. (There is an old package `logger` https://pypi.org/project/logger/ but the most common package is called `logging`.)

jlnav:
> This change has been made in #576. `logger` is possibly external facing... but I think the risk of confusion is near nothing. (There is an old package `logger` https://pypi.org/project/logger/ but the most common package is called `logging`.)

Good point. On the off-chance a libE user is also using the other 'logger', it could be resolved with ``from libensemble import logger as libE_logger``.

A rename of the logger module would technically be a breaking API change then since we intend that part to be user facing. Thoughts on saving that rename for later or getting it over with now? I wouldn't mind either way.

jmlarson1:
@shuds13 As the `logger` is API breaking, what do you think about making a branch for v0.8 that I can make the target of the PR #576 ?

jmlarson1:
As the `logger` rename is API breaking, we can do it now in #608

jlnav:
I believe this has been completely satisfied as of #608 .

PR #576, opened 2021-01-06
Internal/removing libE from filenames

jmlarson1:
Addresses issue #573


jlnav (approved):

jlnav:
@shuds13 and I were thinking that the logger's name could be kept as-is for the upcoming minor release, then it can be changed before the next major release 0.8.0, sorry for the inconvenience. Thoughts?

jmlarson1:
Sounds great. I did my commits very modular for this very reason. I think you can just un-rename the `libE_logger.py` file and undo the commit https://github.com/Libensemble/libensemble/pull/576/commits/3a2a49e7d54efefbe80b566eb0610b9ec5a66083

Issue #571, opened 2021-01-05
Default alloc_f fails with KeyError when using empty persis_info


jlnav:
``persis_info = {}``

```
[0] libensemble.libE (ERROR): Traceback (most recent call last):
  File "/Users/John-Luke/Desktop/libensemble/libensemble/libensemble/libE.py", line 155, in libE_manager
    exit_criteria, persis_info, wcomms)
  File "/Users/John-Luke/Desktop/libensemble/libensemble/libensemble/libE_manager.py", line 88, in manager_main
    result = mgr.run(persis_info)
  File "/Users/John-Luke/Desktop/libensemble/libensemble/libensemble/libE_manager.py", line 475, in run
    persis_info)
  File "/Users/John-Luke/Desktop/libensemble/libensemble/libensemble/libE_manager.py", line 449, in _alloc_work
    output = alloc_f(self.W, H, self.sim_specs, self.gen_specs, self.alloc_specs, persis_info)
  File "/Users/John-Luke/Desktop/libensemble/libensemble/libensemble/alloc_funcs/give_sim_work_first.py", line 84, in give_sim_work_first
    gen_work(Work, i, [], [], persis_info[i])
KeyError: 1
```

- [ ] Fix ``alloc_f`` OR initialize ``persis_info`` with workerIDs in ``libE.py``

jlnav:
Closed by #578

PR #578, opened 2021-01-06
Replace persis_info[i] with persis_info.get(i) in several alloc_f functions

jlnav:
Addresses #571 

An approach to many ``alloc_f`` functions like the default failing with a ``KeyError`` if an empty ``persis_info`` dictionary is passed to ``libE()``. 

This occurs since most of the ``alloc_f``s index ``persis_info`` by workerID using square brackets, producing the error if the workerID keys appended by ``add_unique_random_streams()`` don't exist. Using ``persis_info.get(i)`` instead returns None. This should be fine if no user function requires a ``persis_info`` entry.

Alternative approaches may be best. I considered initiating workerIDs within ``persis_info`` in ``libE()``, but then I think we'd require ``nworkers`` information passed in ``libE_specs`` at runtime. Currently, I think passing ``libE_specs`` (and even ``persis_info``) is supposed to be optional.

### DIFF
--- a/libensemble/alloc_funcs/fast_alloc.py
+++ b/libensemble/alloc_funcs/fast_alloc.py
@@ -28,6 +28,6 @@ def give_sim_work_first(W, H, sim_specs, gen_specs, alloc_specs, persis_info):
             # Give gen work
             persis_info['total_gen_calls'] += 1
             gen_count += 1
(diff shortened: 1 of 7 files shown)


shuds13 (approved):
Looks good.

PR #584, opened 2021-02-24
Removing usage of np.float and np.int

jmlarson1:
`np.float` and `np.int` are deprecated:

https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

I've opened a PR with DFO-LS, which is also affected by this deprecation, which is causing our regression tests to fail (as we error on warnings).

https://github.com/numericalalgorithmsgroup/dfols/pull/13


shuds13 (approved):

jmlarson1:
Even though our regression tests will fail because of DFO-LS, I'm pullling it in.

Issue #587, opened 2021-03-02
libE_stats.txt should give sim_id for sim funcs instead of a count


shuds13:
For sims, change Calc to sim_id.

- This should reflect the naming of output directories for sim_ids (should re-use code), and is more intuitive for users. 
- Will need to check postprocessing scripts still work.
- Gens can still use sequential numbers - also mirroring output directories.
- As before the output will be right aligned at width 5, but will expand if required.

jlnav:
Tentatively closed by #588 ?

PR #588, opened 2021-03-02
Give sim_id in libE_stats.txt

shuds13:
Resolving  #587

### DIFF
--- a/libensemble/output_directory.py
+++ b/libensemble/output_directory.py
@@ -88,7 +88,7 @@ class EnsembleDirectory:
         return any([setting in self.specs for setting in dir_type_keys])
 
     @staticmethod
-    def _extract_H_ranges(Work):
+    def extract_H_ranges(Work):
         """Convert received H_rows into ranges for labeling """
         work_H_rows = Work['libE_info']['H_rows']
         if len(work_H_rows) == 1:
@@ -176,7 +176,7 @@ class EnsembleDirectory:
             self.loc_stack = LocationStack()
 
         if calc_type == EVAL_SIM_TAG:
-            H_rows = self._extract_H_ranges(Work)
+            H_rows = self.extract_H_ranges(Work)
         else:
             H_rows = str(calc_iter[calc_type])
 
--- a/libensemble/tests/unit_tests/test_sim_dir_properties.py
+++ b/libensemble/tests/unit_tests/test_sim_dir_properties.py
@@ -10,7 +10,7 @@ def test_range_single_element():
 
     work = {'H_fields': ['x', 'num_nodes', 'ranks_per_node'],
             'libE_info': {'H_rows': np.array([5]), 'workerID': 1}}
-    assert EnsembleDirectory._extract_H_ranges(work) == '5', \
+    assert EnsembleDirectory.extract_H_ranges(work) == '5', \
         'Failed to correctly parse single H row'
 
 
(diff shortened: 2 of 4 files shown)


shuds13:
@jlnav Do you think the static function _extract_H_ranges could drop the underscore and be public?

jlnav (approved):
Looks good. I may make ``extract_H_ranges()`` a separate internal helper function at some point (not on this PR), so the worker or potentially another module doesn't need to envoke the EnsembleDirectory class for logging purposes.

jlnav:
So, ``extract_H_ranges()`` returns ranges separated by underscores. e.g. the rows 0, 1, 2, 3, 7, 8 returns ``'0-3_7-8'``. This is fine for directories, but for ``libE_stats.txt``, maybe ``0-3, 7-8`` looks better. This adjustment could be done with ``.replace('_', ', ')``, or adding a new parameter to ``extract_H_ranges`` for a separater.

shuds13:
I think it would look better but we would need to check post processing scripts still work, if it had a space in it. Some of the plot scripts read this file. I'm inclined to think its okay as is (and also probably quite rare to have multiple ranges).

shuds13:
Here is  a snippet from libE_stats.txt for test_uniform_sampling_with_different.py
```
Worker     1: Gen no     1: gen Time: 0.000 Start: 2021-03-01 22:22:56.174 End: 2021-03-01 22:22:56.174 Status: Not set
Worker     2: Gen no     1: gen Time: 0.001 Start: 2021-03-01 22:22:56.176 End: 2021-03-01 22:22:56.177 Status: Not set
Worker     3: Gen no     1: gen Time: 0.000 Start: 2021-03-01 22:22:56.179 End: 2021-03-01 22:22:56.179 Status: Not set
Worker     2: sim_id     5: sim Time: 0.808 Start: 2021-03-01 22:22:56.178 End: 2021-03-01 22:22:56.986 Status: Completed
Worker     1: sim_id   0-4: sim Time: 4.051 Start: 2021-03-01 22:22:56.176 End: 2021-03-01 22:23:00.227 Status: Completed
Worker     1: sim_id     6: sim Time: 0.406 Start: 2021-03-01 22:23:00.228 End: 2021-03-01 22:23:00.634 Status: Completed
Worker     1: Gen no     2: gen Time: 0.001 Start: 2021-03-01 22:23:00.638 End: 2021-03-01 22:23:00.639 Status: Not set
Worker     1: sim_id     7: sim Time: 0.806 Start: 2021-03-01 22:23:00.640 End: 2021-03-01 22:23:01.446 Status: Completed
```

The gen is just enumerating still on that worker which I don't love, but we don't know what gens have been run on other workers ( I suppose you could if the alloc func kept a count and put it in libE_info). 
Anything better than 'gen no'
Maybe "sim Time" or "gen Time" can just be Time.

jlnav:
> The gen is just enumerating still on that worker which I don't love, but we don't know what gens have been run on other workers ( I suppose you could if the alloc func kept a count and put it in libE_info).

I'm sure we'd prefer to make this change without modifying all our allocation functions. Maybe ``gen_work()`` could either instantiate or update a count and append it to libE_info

> Anything better than 'gen no'

If anything, I like that it's the same character length as "sim_id". "gen_id" would be incorrect of course.

> Maybe "sim Time" or "gen Time" can just be Time.

Agreed.

I can open an issue for additional libE_stats improvements, unless we'd like to incorporate some of them into this PR

PR #592, opened 2021-03-23
Bugfix/fix error handling

shuds13:
Make exception handling more robust.

Catch exceptions in manager to ensure they get logged when worker
exceptions occur in the finally block.

Exception handling remains in libE to ensure exceptions in the manager
finally block are handled and abort/dump happens.

Final raise is disabled, this should be reviewed.

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -160,16 +160,16 @@ def manager(wcomms, sim_specs, gen_specs, exit_criteria, persis_info,
             raise
         except WorkerException as e:
             report_worker_exc(e)
-            raise
-        except Exception:
+            raise LoggedException(e.args[0], e.args[1]) from None
+        except Exception as e:
             logger.error(traceback.format_exc())
-            raise
-    except Exception:
+            raise LoggedException(e.args) from None
+    except Exception as e:
         exit_flag = 1  # Only exits if no abort/raise
         _dump_on_abort(hist, persis_info, save_H=save_H)
         if libE_specs.get('abort_on_exception', True) and on_abort is not None:
             on_abort()
-        raise LoggedException('See specific error above and in ensemble.log') from None
+        raise LoggedException(*e.args, 'See error details above and in ensemble.log') from None
     else:
         logger.debug("Manager exiting")
         logger.debug("Exiting with {} workers.".format(len(wcomms)))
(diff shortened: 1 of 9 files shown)


jmlarson1:
@shuds13 It looks like the expected exception raising in `tests/unit_tests/test_libE_main.py` conflicts with this branch.

In `test_exception_raising_manager_no_abort`, the `MPISendException` is raised, but I'm not sure what more `pytest` wants inside of a block:

`    with pytest.raises(MPISendException):`

shuds13:
> @shuds13 It looks like the expected exception raising in `tests/unit_tests/test_libE_main.py` conflicts with this branch.
> 
> In `test_exception_raising_manager_no_abort`, the `MPISendException` is raised, but I'm not sure what more `pytest` wants inside of a block:
> 
> ` with pytest.raises(MPISendException):`

This is because I commented out the final raise. We could instead just check the exit flag is 1. But I want to get opinions on whether we should raise the error up or simply return the 1 exit code.

jmlarson1:
I can see pros/cons with both raising an error or returning. As a default, I'm not sure how much I check the error codes, so my default would be to expect the error to be raised.

shuds13:
Yes I think I agree that raising is best.

* With just exit_flag - you may just continue without meaning to. You may continue into another loop of libE() when you dont want to (you might want to and you can capture the raised exception - but more likely you dont).

* With raise - you have repeated reporting of exceptions (also minor, but a worker exception from the finally block will not be formatted correctly).

* With sys.exit - dont have the option of capturing/handling the error at the user level.

* With raise LoggedException from None. Reported on but without repeating whole stack - could lose specific information if handling error from user script, but I think that will mainly want to know simply whether or not there was an error.

jmlarson1:
Great. Let's `raise` the error then.

shuds13:
Yes I think I agree that raising is best.

* With just exit_flag - you may just continue without meaning to. You may continue into another loop of libE() when you dont want to (you might want to and you can capture the raised exception - but more likely you dont).

* With raise - you have repeated reporting of exceptions (also minor, but a worker exception from the finally block will not be formatted correctly).

* With sys.exit - dont have the option of capturing/handling the error at the user level.

* With raise LoggedException from None. Reported on but without repeating whole stack - could lose specific information if handling error from user script, but I think that will mainly want to know simply whether or not there was an error.

PR #594, opened 2021-03-25
Hugely simplify sim_dir exception test

jlnav:
Fixes my silly design choice of having this test rely on output from a previous calc_dir test


jmlarson1:
Looks good to me

PR #595, opened 2021-03-29
Collab/cwp calibration

shuds13:
Merge in surmise calibration branch. 

This branch includes implementation of a mechanism for cancellation of previously issued evaluations, and sending kill signals to running evaluations that have been cancelled. An active receive worker mode is added allowing irregular communication patterns between manager and worker.

Two examples are provided using the surmise library for a generator function. A tutorial is also included.

Also a submit is added to the base Executor class that sub-processes a serial application locally.


mosesyhc (approved):
I have pushed some minor edits to the comments for surmise gen. Otherwise it looks good to me.

jlnav (approved):
I resolved the ``ResourceWarning`` on the Python 3.8+ jobs by opening the Task's ``stdout`` and ``stderr`` in a ``with`` statement, then performing the launch in the block. This way those files are closed once the launch concludes.

This is the same solution to that "deja vu" with the MPIExecutor, so both modules have very similar looking launch blocks.

Also, maybe the ``persistent_surmise`` example page should say that the tutorial uses the gen as an example function. The tutorial (in it's current form) doesn't really teach users how to use the generator.

Looks great!

mosesyhc:
> I resolved the `ResourceWarning` on the Python 3.8+ jobs by opening the Task's `stdout` and `stderr` in a `with` statement, then performing the launch in the block. This way those files are closed once the launch concludes.
> 
> This is the same solution to that "deja vu" with the MPIExecutor, so both modules have very similar looking launch blocks.
> 
> Also, maybe the `persistent_surmise` example page should say that the tutorial uses the gen as an example function. The tutorial (in it's current form) doesn't really teach users how to use the generator.
> 
> Looks great!

I agree with the last comment on "using the gen as an example function" (for now). On the surmise package end, we are completing the documentation for the method and some usage examples. We may point to that in the tutorial as it becomes available.

docs/tutorials/calib_cancel_tutorial.rst
```diff
@@ -0,0 +1,316 @@
...
+..
+..     def subproc_borehole_func(H, subp_opts, libE_info):
+..         sim_id = libE_info['H_rows'][0]
+..         H_o = np.zeros(H.shape[0], dtype=sim_specs['out'])
+..         ...
+..         exctr = Executor.executor
+..         task = exctr.submit(app_name='borehole', app_args=args, stdout='out.txt', stderr='err.txt')
+..         calc_status = polling_loop(exctr, task, sim_id)
+..
+.. asdf ``polling_loop()`` resembles the following::
```
jmlarson1, line 239: Is the `asdf` a keyboard smash? :-)

jmlarson1 (approved):

docs/tutorials/calib_cancel_tutorial.rst
shuds13: I assume so. Fixed.

PR #596, opened 2021-04-16
Feature/general alloc support cancel

shuds13:
Ensuring cancel is supported by all relevant allocation functions. Also updates docs.


shuds13:
@jmlarson1 
I have now incorporated _cancel_requested_ into all alloc_funcs except for `fast_alloc_and_pausing.py`.  As this has its own sort of cancellation I'm not sure if you would ever want to combine it? Alternatively the _cancel_requested_ flag could be set in the allocation function to implement the cancellations. Let me know what you think?

I've also added an alloc_support function `all_returned` for when you must wait on all points to return before giving back to the gen as its slightly non-trivial considering cancellations. 

i could also add a _skip_cancelled_ for these ines that I put in the fast_alloc functions. If its worth it?
```
while H[persis_info['next_to_give']]['cancel_requested']:
            persis_info['next_to_give'] += 1
```

jmlarson1:
I like the `cancel_requested` field. I think it's okay to leave an example of an alloc_func (e.g., `fast_alloc_and_pausing.py`) that handles such a feature itself. 

I don't think that we want the user function to set the `cancel_requested` flag. I think of that as a libE field not user field. 

I think `all_returned` is good, but what happens if the same point is given multiple times? 

I'm not sure I follow what a `skip_cancelled` field would be? Is that a libE_spec that, if true, skips cancelled?

shuds13:
I'm okay to keep `fast_alloc_and_pausing` as it is - maybe I will say in docstring does not support cancel_requested. It is a field that can be set by the gen, but it would be a bit different setting it here for sure.

If the points are called multiple times we have to change the code to use the new system - e.g., something like returned < given. Having a function might help us make the change once for all routines?

The skip_cancelled function would just hold the loop that I put in the fast_alloc functions. It just iterates to the next non_cancelled point (or the end of H). I appreciate the non-persistent functions are probably unlikely to use cancellations but it seems like general support would be expected.

jmlarson1:
Instead of "does not support", perhaps "does not make use of"? 

We'll have to address how to interpret `cancel_requested` when we allow for points to be given multiple times. (Or maybe we just don't allow both to be used.)

shuds13:
> We'll have to address how to interpret `cancel_requested` when we allow for points to be given multiple times. (Or maybe we just don't allow both to be used.)

Is it likely that cancel_requested would be set and then re-given, in which case the gen could reset it to False. 

One difference you might want in that case, is to declare cancel_requested in gen_specs['out'] as resetting it to False would probably be combined with a regular gen_out communication.

shuds13:
In give_sim_work_first we have this:

```python
            if 'in' in gen_specs and len(gen_specs['in']):
                gen_work(Work, i, gen_specs['in'], range(len(H)), persis_info.get(i))
            else:
                gen_work(Work, i, [], [], persis_info.get(i))
```

In the fast_alloc variants we just have:
```python
    gen_work(Work, i, gen_specs['in'], range(len(H)), persis_info.get(i))
```

I am assuming there is not a reason for these not have the same option of no gen_specs['in'], if for example just random sampling. So I am inclined to add the top block to these.

Also maybe would be better to have one function call eg.

```python
            gen_in = gen_specs.get('in', [])
            return_rows = range(len(H)) if gen_in else []
            gen_work(Work, i, gen_in, return_rows, persis_info.get(i))
```

jmlarson1:
I think that making all of the alloc_funcs more robust (i.e., working when they don't have all input parameters ) is good. 

I'm fine the "one function call"  form you propose

jlnav (approved):
Looks good. Would it make sense for the new cancellation logic:

```
while persis_info['next_to_give'] < len(H) and H[persis_info['next_to_give']]['cancel_requested']:
    persis_info['next_to_give'] += 1
```

to be made into an additional helper function in alloc_support? e.g.:

`persis_info = skip_cancelled_points(persis_info, H)`

jmlarson1:
Is there a reason the CI tests aren't passing?

shuds13:
> Is there a reason the CI tests aren't passing?

There is a Surmise dependency issue. We may just disable those tests for now, and re-enable them once fixed.,

shuds13:
@jlnav 

I did think about that, it probably should. I was a bit unsure given the simplicity of it, whether to just have the logic show in place. Secondly, the helper function knowing the 'next_to_give' in persis info - was troubling me a bit. The alternative would be to make it functional e.g. persis_info['next_to_give'] = skip_cancelled_points(persis_info['next_to_give'], H). But not too keen on that either.

jmlarson1 (approved):

PR #600, opened 2021-04-27
Feature/uneven workers nodes

### LINKED ISSUE #591: sub node worker assignment
I ran into an error trying to assign 8 workers across 3 nodes and realized that it appears that currently it appears that when trying to configure a run with `workers > nodes` that `workers % nodes` must be 0. 

This appears to come from the node assignment:
https://github.com/Libensemble/libensemble/blob/0d699f404c0a93455dbe0c25fe230c086bee7831/libensemble/resources/resources.py#L335-L338

Would it be feasible to extend this to handle cases where `workers % nodes != 0` ? 
I tried a naive implementation that extends the `global_nodelist` based on this remainder and it seemed to fix the issue for my case. Perhaps there are broader implications for such a change though?

shuds13:
Support uneven distribution of workers to nodes.

Resolves #591

Re-did PR due to merging in from develop.


shuds13:
> Looks good. Can you help me understand what is tested by the two `subnode_uneven` that isn't tested in the `supernode_uneven` regression tests? It would seem like "multiple and uneven nodes per worker" covers the case of "uneven workers per node".

The supernode case has a full number of nodes for each worker. The subnode case, has multiple workers on each node. Does this answer?

I did think I should maybe put the _runline_ tests in a different directory as they are kind of artificial tests.

jmlarson1 (approved):
Looks good

jlnav (approved):

Issue #572, opened 2021-01-05
Implement helper function or executor option for generic task status loop


jlnav:
Users who simply want to wait until an app concludes before evaluation, leave control up to ``manager_poll()``, or predictably set ``calc_status`` could take advantage of a simple looping function or routine within the executor.

Alternatively, ``calc_status`` could be set and ``manager_poll()`` called periodically within ``task.wait()``

jlnav:
Closed by #612

PR #612, opened 2021-05-07
Executor.polling_loop(task, args*) interface

jlnav:
Addresses #572 

Implements generic blocking polling_loop method to Task class. Potentially useful for users who want to simply run an app to completion, timeout, or kill and receive a presumptive calc_status.

Potential TODOs:

~~- [ ] Investigate non-blocking or poll_all_tasks approach.~~

~~- [ ] Deprecate Executor.manager_poll?~~

- [x] Also test with surmise killsims scenario

5/13 Updates:

- [x] Instead considering `exctr.polling_loop(list_of_tasks, *args)` interface.


libensemble/executors/executor.py
```diff
@@ -435,6 +439,68 @@ def manager_poll(self):
...
+            task.poll()
+
+            if poll_manager:
+                man_signal = self.manager_poll()
+                if self.manager_signal != 'none':
+                    task.kill()
+                    calc_status = man_signal
+                    break
+
+            if timeout > 0 and task.runtime > timeout:
```
shuds13, line 487: Maybe timeout should default to None, rather than 0 ?? Open to opinions.

PR #615, opened 2021-05-10
Experimental persistent sim test

shuds13:


shuds13:
Still need to re-visit gen_support - should it be worker_support persis_support or whatever...
- and handle the conflicts.
-

PR #618, opened 2021-05-18
Testing use_persis_return capabilities

jmlarson1:


libensemble/gen_funcs/persistent_deap_nsga2.py
```diff
@@ -68,7 +68,7 @@ def evaluate_pop(g, deap_object, Out, comm):
         Out['generation'][index] = g
     # Sending work to sim_f, which is defined in main call script
     # A fitness value will be returned in calc_in
-    tag, Work, calc_in = sendrecv_mgr_worker_msg(comm, Out)
+    tag, Work, calc_in = sendrecv_mgr_worker_msg(comm, Out[['individual', 'generation']])
```
shuds13, line 71: Do you need to specify these fields here? Last points will be false anyway?

jmlarson1: The reason is more to show/test that something such as `'last_points'` is never communicated from the persistent user script (only from the worker after the persistent worker has ended).

PR #630, opened 2021-06-15
Experimental DeepDriveMD inspired workflow

jlnav:
An experimental libEnsemble workflow that coordinates each of the Molecular Dynamics, Machine Learning, and other data processing components of DeepDriveMD in a similar way to DeepDriveMD. A persistent gen initiates an initial ensemble of MD simulations, then submits a series of ML and data processing apps on the results. The output of these apps informs future ensembles of MD simulations. The purpose of this test is as an example workflow for evaluating the efficacy of libEnsemble coordinating MD/ML workflows.

Currently, only a single ML app is launched for each ensemble of MD sims; I want to address this in future updates. Currently performing testing on Swing.

Please leave any comments or suggestions. I don't plan on possibly merging this in until we've met with the DeepDriveMD devs. Thanks!


jmlarson1 (commented):
Looks good to me. Perhaps consider using `wget` to get the large data files.

shuds13 (approved):

jmlarson1 (commented):
Is there a reason why the `gen` and `sim` are not included in the normal `gen` and `sim` directories?

jlnav:
> Is there a reason why the `gen` and `sim` are not included in the normal `gen` and `sim` directories?

I thought it acceptable since the gen and sim are dedicated to this test case, and no other. At this time, I don't expect us (or other users) to import and use them in separate tests or use-cases. _Although_, they do make nice examples for including in the docs or symlinking to in the ``examples`` directory.

The other scaling tests are also the same way, but I'd be open to adjusting this.

jmlarson1 (approved):

PR #633, opened 2021-06-21
Feature/additional termtest check

jmlarson1:


shuds13:
Not sure why we need another term test here, unless there has been significant time since the last one (for a timeout).

jmlarson1:
I thought it odd to call the allocation function if the last receive_from_worker resulted in the termination test being satisfied. I can remove that if you'd like.

shuds13:
That makes sense.

shuds13 (approved):

PR #637, opened 2021-06-23
Allowing to users to tell the manager to not check whether to cancel …

jmlarson1:
…points

### DIFF
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -71,7 +71,12 @@ Specifications for libEnsemble::
         'authkey' [string]:
             TCP Only: Authkey
         'safe_mode' [boolean]:
-            Prevents user functions from overwritting protected libE fields.
+            Prevents user functions from overwritting protected libE fields, but requires
+            moderate overhead.
+            Default: True
+        'check_cancel' [boolean]:
+            Will the user functions possibly request cancelling a point. If False, the manager 
+            avoid the moderate overhead of seeing what points need to be cancelled. 
             Default: True
         'use_persis_return' [boolean]:
             Adds persistent function H return to managers history array.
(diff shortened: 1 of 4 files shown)


shuds13 (commented):
I'm wondering if check_cancel is best term.

Its really about whether it sends out a kill signal to cancelled sims. Maybe more like 'kill_cancelled', or something better?

jmlarson1:
@shuds13 The intention is to not even check if any user function has requested a cancel if `check_cancel` is False. For fast sim_f evaluation times and long runs, the check
````
kill_sim = self.hist.H['given'] & self.hist.H['cancel_requested'] \
                & ~self.hist.H['returned'] & ~self.hist.H['kill_sent']
````
becomes a major bottleneck in the manager. 

Related (but a different PR) would be finding is a more intelligent way of doing that check. For a `sim_id` that is `returned`, we won't be sending a kill.

shuds13:
I planned to include the sliding window approach in this check (and anything that did a whole of H filter), when possible. But even still that whole block of code is deciding whether to send a kill. If a sim is cancelled before being sent out, it will not be sent by the allocation function, and so I don't want to confuse with that check.

jmlarson1:
> If a sim is cancelled before being sent out, it will not be sent by the allocation function, and so I don't want to confuse with that check.

Wouldn't it have to be the user function that cancels a sim before being sent out?

shuds13:
Yes, user function sets cancel_requested.

shuds13:
I also wonder if there should be a compound option to make everything as fast as possible in this respect - eg. dont check safe mode, dont check/send kills

jmlarson1:
So I'd like to make a boolean that would tell the manager not to check if any cancel has been requested. I thought `libE_specs['check_cancel']` was an okay name. Any suggestions for something better?

shuds13:
I suggest kill_cancelled_sims
I've add this to the terminology wiki https://github.com/Libensemble/libensemble/wiki/Terminology-structure-decisions

shuds13 (approved):

Issue #609, opened 2021-05-06
Construct Class for Persistent Worker support functions


jlnav:
We're considering moving Persistent Gen support functions to a helper class, that can contain (among other attributes) a comm attribute. Users can import and instantiate this class, or extend for their own needs.

PR #639, opened 2021-06-25
Experimental persistent worker support class

jlnav:
Addresses #609 

Opening this PR primarily to eye-test how such a class could look. My first impression is lines-of-code are not necessarily reduced, but tracking statistics, logging, and adding additional functions or tools should be easier. Comments and suggestions _hugely_ appreciated.


shuds13:
I'll try this out with #615

PR #644, opened 2021-07-27
added batched construction for Tasmanian

mkstoyanov:
Adding capabilities for static grid construction and batch refinement algorithms. The main contribution is the modification of the refinement loop, `while grid.getNumNeeded() > 0:` and performing refinement operations after the manager reports back each set of points.

### DIFF
--- a/libensemble/gen_funcs/persistent_tasmanian.py
+++ b/libensemble/gen_funcs/persistent_tasmanian.py
@@ -16,7 +16,7 @@ def sparse_grid_batched(H, persis_info, gen_specs, libE_info):
 
     """
     U = gen_specs['user']
-    grid = U['grid']
+    grid = U['tasmanian_init']()  # initialize the grid
     allowed_refinements = ['setAnisotropicRefinement', 'setSurplusRefinement', 'none']
     assert 'refinement' in U and U['refinement'] in allowed_refinements, \
         "Must provide a gen_specs['user']['refinement'] in: {}".format(allowed_refinements)
--- a/libensemble/gen_funcs/persistent_tasmanian.py
+++ b/libensemble/gen_funcs/persistent_tasmanian.py
@@ -4,9 +4,9 @@ A persistent generator using the uncertainty quantification capabilities in
 """
 
 import numpy as np
-import Tasmanian
 from libensemble.message_numbers import STOP_TAG, PERSIS_STOP, FINISHED_PERSISTENT_GEN_TAG
 from libensemble.tools.gen_support import sendrecv_mgr_worker_msg
+
 
 def sparse_grid_batched(H, persis_info, gen_specs, libE_info):
     """
@@ -17,6 +17,9 @@ def sparse_grid_batched(H, persis_info, gen_specs, libE_info):
     """
     U = gen_specs['user']
     grid = U['grid']
+    allowed_refinements = ['setAnisotropicRefinement', 'setSurplusRefinement', 'none']
+    assert 'refinement' in U and U['refinement'] in allowed_refinements, \
+        "Must provide a gen_specs['user']['refinement'] in: {}".format(allowed_refinements)
 
     while grid.getNumNeeded() > 0:
         aPoints = grid.getNeededPoints()
@@ -52,6 +55,6 @@ def sparse_grid_batched(H, persis_info, gen_specs, libE_info):
                 assert 'sCriteria' in U
                 grid.setSurplusRefinement(U['fTolerance'], U['iOutput'], U['sCriteria'])
             else:
-                pass # called with refinement = none (or some typo)
+                pass  # called with refinement = none (or some typo)
 
     return H0, persis_info, FINISHED_PERSISTENT_GEN_TAG
--- a/libensemble/gen_funcs/persistent_tasmanian.py
+++ b/libensemble/gen_funcs/persistent_tasmanian.py
@@ -43,16 +43,15 @@ def sparse_grid_batched(H, persis_info, gen_specs, libE_info):
             grid.write(U['tasmanian_checkpoint_file'])
 
         # set refinement, using user['refinement'] to pick the refinement strategy
-        if 'refinement' in U:
-            if U['refinement'] == 'setAnisotropicRefinement':
-                assert 'sType' in U
-                assert 'iMinGrowth' in U
-                assert 'iOutput' in U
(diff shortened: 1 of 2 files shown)


jmlarson1:
Thank you for this PR. 

1. The grid is constructed in the `gen_f` but then the points must be communicated to the manager (to then be evaluated by the workers). Is there a reason that this wouldn't be sufficient? (If the number of points in the grid is very large but the simulation time corresponding to each point in the grid is small, then communication costs can be reduced by sending groups of points to be evaluated.)
2. I've added a check that the `refinement` is one of three allowable strings. Feel free to adjust as needed.

mkstoyanov:
My concern is that `gen_f` is very cheap and I don't want to waste an MPI rank that would be almost entirely idle.

But I guess that can be resolved with rank mapping or something of this sort. I'll let you know if I encounter any issues.

mkstoyanov:
@jmlarson1, thanks for adding the error chacking!

jmlarson1:
> My concern is that `gen_f` is very cheap and I don't want to waste an MPI rank that would be almost entirely idle.
> 
> But I guess that can be resolved with rank mapping or something of this sort. I'll let you know if I encounter any issues.

Yes, if the simulation computational cost is much higher than the gen_f, we would have the manager, the gen_f and the first sim_f all share resources.

mkstoyanov:
I think I need to add a test for the surplus refinement, I'm only testing the anisotropic one. Will do today.

jmlarson1:
@mkstoyanov 

For the `run==1` instance in the regression test, I am getting some unexpected hanging with the line 

`            grid.setAnisotropicRefinement(U['sType'], U['iMinGrowth'], U['iOutput'])
`

in the file `gen_funcs/persistent_tasmanian.py` when I run libEnsemble in `local` mode. It does not appear when running in MPI. Also, the issue occurs on multiple systems (and in the Gitub-CI actions). 

Is there something about that process that might clash (or cause a hang) when used conjunction with the python `multiprocessing` package?

For the `run==2` regression test, a hang occurs even earlier: at the line 
`        aPoints = grid.getNeededPoints()
`
of `gen_funcs/persistent_tasmanian.py`


Are you able to run in the following manner? 

` python3 test_persistent_tasmanian.py --comms local --nworkers 4`

mkstoyanov:
It's  the OpenMP, the command that you gave me hangs, but the following works:
```
OMP_NUM_THREADS=1 python3 test_persistent_tasmanian.py --comms local --nworkers 4
```

In the case of `run == 1`, the `getNeededPoints()` and `loadNeededPoints()` methods don't use OpenMP (methods are trivial and cheap). The refinement method uses multi-threading. In the case of `run == 2` and the local polynomial grid, the `getNeededPoints()` method is non-trivial and uses OpenMP. Both cases hang on the first call that uses OpenMP.

Setting `OMP_NUM_THREADS=1` seems to fix the issue. OpenMP can be disabled altogether for Tasmanian in CMake, but not in the pip installer. I'm not sure if there is some environment variable that can make things work when Tasmanian C++ code is running OpenMP through a Python ctypes interface and then sitting underneath the libEnsemble parallelization.

(before I made the PR I tested this with OpenMPI and `mpirun` and there was no issue)

shuds13:
I think this is the problem: https://stackoverflow.com/questions/59251966/why-does-my-function-in-a-shared-library-which-uses-openmp-hang-when-called-from

But if you do OMP_NUM_THREADS=1 on command line
but then in gen do: os.environ["OMP_NUM_THREADS"] = "2"
then that does not hang, atleast

mkstoyanov:
@shuds13 the issue seems in the order of calls to the parallel engines.

Right now, we have OpenMP call in the Tasmanian `make<some>Grid()`, followed by libEnsemble call to do the sampling, and in the sampling we are calling Tasmanian methods with OpenMP again.

The correct way is to make all Tasmanian OpenMP calls AFTER libEnsemble has initialized the parallel workers. Basically, we need to call the Tasmanian `make<some>Grid()` method inside the `gen_f()` call.

We can do that similar to the `set<some>Refinement()` but that means we have to pass in lots and lots of parameters, translating all the possible grid initialization calls to Tasmanian into `user` parameters in libEnsemble will be tedious to write and maintain.

A better approach is to make another method similar to `gen_f` and `sim_f`, call it `tasmanian_init`. The method will be written by the user and it will return a `TasmanianSparseGrid` object.

@shuds13 @jmlarson1 are you guys OK with that solution? I can update the pull request today.

mkstoyanov:
I was wrong!

OpenMP is initialized in the `import Tasmanian` line, not the `make<some>Grid()` command. Delaying the make command does not help. The following does:

```
import os
has_omp_num_threads = "OMP_NUM_THREADS" in os.environ
if has_omp_num_threads:
    omp_num_threads = os.environ["OMP_NUM_THREADS"]
    os.environ["OMP_NUM_THREADS"] = "1"  # temporarily disable OpenMP
import Tasmanian
if has_omp_num_threads:
    os.environ["OMP_NUM_THREADS"] = omp_num_threads
```

Not sure if this is a good way to do things.

jmlarson1:
Can we do this inside of the `gen_f` script? From our point of view, it seems like we could have the `import Tasmanian` inside of the `gen_f` and move the snippet 

    if run < 2:
        # tests the cases of anisotropic refinement and no refinement
        grid = Tasmanian.makeGlobalGrid(num_dimensions, 1, 6, "iptotal", "clenshaw-curtis")
    else:
        # tests the surplus refinement
        grid = Tasmanian.makeLocalPolynomialGrid(num_dimensions, 1, 3)

to within the `gen_f`? Then the calling script could declare `gen_specs['user']['grid_args'] = (num_dimensions, 1, 6, "iptotal", "clenshaw-curtis")`

mkstoyanov:
@jmlarson1  In principle, we could do that. However, the `make<some>Grid()` can have 5 different variants (Sequence, Wavelet, Fourier in addition to what we have now) and putting all of those inside libEnsemble and maintaining any new ones will make it messy. I want to find a more elegant solution.

The `tasmanian_init` idea almost works if I just put the `import Tasmanian` clause inside the method. I'm running into issues with the sanity check. 

Run 0 works fine, but then I `import Tasmanian` again to do the sanity check and then followup calls to libEnsemble hang. I'm still playing with this.

jmlarson1:
Thank you very much for your help @mkstoyanov 

I think it looks great. Please let us know if you (or collaborators) end up developing any additions to the Tasmanian gen_f.

PR #653, opened 2021-08-27
Tentative Refactor - Move incomplete/deprecated/externally supported user funcs to separate repo?

jlnav:
We've discussed doing this, based on several considerations:

1. Example user functions take up a huge part of the repo. This isn't necessarily a problem (I think it's great!), but it looks a bit strange when GitHub states our Python package is ~45% Fortran, for example.
2. It's probably not best to distribute code we're not prepared to support/troubleshoot ourselves (which is why I think APOSMM and many of the scaling and/or regression tests can stay in).

   - If a dependency can be easily ``pip`` or ``conda`` installed, that's probably okay?

3. Potential for dramatically slimming down and simplifying testing

For _starters_, I'm removing code on this feature branch. Once a consensus is made, I'll migrate files from ``develop`` into a new repo before pulling this in.

Feel free to contribute directly to this branch or branch off yourself and open PRs into this one. Thanks!


jlnav:
Closing until we've completed a tentative set of principles for the new repo, and we've determined what needs to be moved.

Issue #611, opened 2021-05-06
Split or categorize regression tests by if they require additional dependencies


jlnav:
This could be accomplished by splitting up the regression tests into separate directories or adding additional run-tests.sh parse-able labels to relevant tests. The idea is users should be able to easily run a selection of regression tests to verify libE installation without installing additional dependencies.

jlnav:
Largely satisfied by #654

PR #654, opened 2021-08-27
Feature - Adds filter label for "extra" regression tests

jlnav:
Addresses #611 

Adds label to certain regression tests that excludes them from normal ``./run-tests.sh`` runs. I tried to pick those that have external dependencies that most users (or new users) don't have following a fresh install of libEnsemble. The idea is, following ``git clone ...libensemble.git`` that users can hopefully verify their installation with a quick ``./run-tests.sh``.

For our CI, and for those who want to run the entire regression test suite, the new ``./run-tests.sh -e`` option is available.

Will changes also be necessary for unit tests? I haven't decided yet, but would appreciate feedback on that or anything else. Thanks!

### DIFF
--- a/.github/workflows/libE-ci.yml
+++ b/.github/workflows/libE-ci.yml
@@ -111,14 +111,14 @@ jobs:
         - name: Run tests, Ubuntu
           if: matrix.os == 'ubuntu-16.04'
           run: |
-           ./libensemble/tests/run-tests.sh -A "-W error" -z -${{ matrix.comms-type }}
+           ./libensemble/tests/run-tests.sh -e -A "-W error" -z -${{ matrix.comms-type }}
 
         - name: Run tests, macOS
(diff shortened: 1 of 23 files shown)


jlnav:
I've also noticed that scipy and mpi4py are currently "required" to run the following unit tests:

- test_allocation_funcs.py
- test_libE_main.py
- test_manager_main.py
- test_mpi4py.py
- test_persistent_aposmm.py
- test_scipy.py

Obviously I'd want to keep running the first three, and would probably prefer that they're "core" tests that are runnable after a pip-install of libE. I'm considering exploring adjusting those tests to have MPI and non-MPI versions that can run depending on if mpi4py is found.

As for test_persistent_aposmm.py, this is potentially getting moved to the new repo alongside APOSMM, correct?

jmlarson1 (approved):

PR #659, opened 2021-09-14
Refactor/proposed api changes from wiki

jlnav:
Proposed API renames directly from: https://github.com/Libensemble/libensemble/wiki/Expected-changes-for-libEnsemble-v0.8.0

We've discussed that the time to do such changes is right before a major release. Don't panic from all the changes! We've wanted to do these for a while, and the tests pass locally. The key will be to point users to the wiki or a doc describing our changes, but hopefully all they'll need to do is find-and-replace.


jmlarson1 (approved):

shuds13 (commented):
Theres a lot of conflicts with adaptive_workers here, so I'd hold off on this until that is pulled in. And on anything that touches a lot of files.

jmlarson1 (approved):

shuds13 (approved):

PR #661, opened 2021-09-16
Proposed changes to make alloc support more intuitive

jmlarson1:
This is what I was trying to describe in the talk today. I like the current level of `try... except... ` content in the example alloc_f.

### DIFF
--- a/libensemble/alloc_funcs/start_only_persistent.py
+++ b/libensemble/alloc_funcs/start_only_persistent.py
@@ -69,7 +69,7 @@ def only_persistent_gens(W, H, sim_specs, gen_specs, alloc_specs, persis_info):
         gen_inds = (H['gen_worker'] == wid)
         returned_but_not_given = np.logical_and.reduce((H['returned'], ~H['given_back'], gen_inds))
         if np.any(returned_but_not_given):
-            if async_return or support.all_returned(gen_inds):
+            if async_return or support.all_returned(H, gen_inds):
                 point_ids = np.where(returned_but_not_given)[0]
                 support.gen_work(Work, wid, gen_return_fields, point_ids, persis_info.get(wid),
                                  persistent=True, active_recv=active_recv_gen)
@@ -83,9 +83,9 @@ def only_persistent_gens(W, H, sim_specs, gen_specs, alloc_specs, persis_info):
         if not np.any(points_to_evaluate):
             break
 
-        sim_ids_to_send = support.points_by_priority(points_avail=points_to_evaluate, batch=batch_give)
+        sim_ids_to_send = support.points_by_priority(H, points_avail=points_to_evaluate, batch=batch_give)
         try:
-            support.sim_work(Work, wid, sim_specs['in'], sim_ids_to_send, persis_info.get(wid))
+            support.sim_work(Work, wid, sim_specs['in'], sim_ids_to_send, persis_info.get(wid), H)
         except InsufficientFreeResources:
             break
 
--- a/libensemble/tools/alloc_support.py
+++ b/libensemble/tools/alloc_support.py
@@ -38,9 +38,8 @@ class AllocSupport:
         :param user_scheduler: Optional, A user supplied ``user_scheduler`` object.
         """
         self.W = W
-        self.H = H
         self.persis_info = persis_info
-        self.manage_resources = 'resource_sets' in H.dtype.names
+        self.manage_resources = sched_opts.get('use_resource_sets', False) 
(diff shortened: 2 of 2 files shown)


jlnav:
As we discussed, I don't mind the changes for ``points_by_priority()``, ``all_returned()``, etc. I'll update this with additional resource considerations as I encounter them...

Two points I've thought about, not necessarily related to resources:

1. Now that we're considering passing each of ``H``, ``H_fields`` and ``H_rows`` into ``sim_work()`` and ``gen_work()``, would it be a bad idea to combine those? e.g.:

   `sim_work(self, Work, wid, H_fields, H_rows, persis_info, H=None, **libE_info)`

   could be:

   ``sim_work(self, Work, wid, H_sel, persis_info, **libE_info)``

   where ``H_sel = H[H_fields][H_rows]`` in the ``alloc_f``.

   Would this simplify ``_check_work_order()`` and ``_send_work_order()`` in the manager?

2. Passing ``Work`` into ``sim_work()`` and ``gen_work()`` to be updated in-place based ``wid`` keys is very C-like when those functions could return ``Work`` dictionary entries instead, e.g.:

   ``Work[wid] = sim_work(wid, H_sel, persis_info.get(wid))``

shuds13:
Okay, taking these points one at a time.

For the modified code, I guess its okay just to pass H into sim_work. I'm not sure about the use_resource_sets option. Currently it would only make sense if resource_sets is a gen_out - otherwise  its not in H - unless we use a .get(resource_sets,1). But if the users has already had to think about it, to add it as a gen_out - do we need an option also. Perhaps if you want a quick way to turn it on and off? But then what is the default? It would also become an inactive option is disable_resource_manger=True.

jmlarson1:
We can (and should) assert that `resource_sets` is a `gen_out` if `use_resource_sets` is True.

shuds13:
But if resource_sets is a gen_specs['out'], then you also need to add use_resource_sets=True ?

The only reason I can see, and maybe its a good reason, is if you want to have a gen that sets resource_sets, but say you want to ignore it.

jmlarson1:
> But if resource_sets is a gen_specs['out'], then you also need to add use_resource_sets=True ?

I think of this as the opposite. If `use_resource_sets` is True, the user must supply them. If (by happenstance) they have a `gen_specs['out']` field that is called `resource_sets`, but `use_resources_sets` is False (or unset), then we'd ignore them.

shuds13:
Okay, but I could compare with with 'priority' where we dont have a use_priority option, we just look if the field is there.

jlnav:
> I think of this as the opposite. If `use_resource_sets` is True, the user must supply them. If (by happenstance) they have a `gen_specs['out']` field that is called `resource_sets`, but `use_resources_sets` is False (or unset), then we'd ignore them.

Hm, yeah, I'd think a user that explicitly specifies ``'resource_sets'`` as a gen_out wants to take advantage, automatically, of resource set features

shuds13:
>     1. `sim_work(self, Work, wid, H_sel, persis_info, **libE_info)`
>        where `H_sel = H[H_fields][H_rows]` in the `alloc_f`.

If I'm reading this right, H_sel would contain the actual data. I guess its a view but  it would need to go with a single send, which needs revisiting comms. So I would keep work as metadata, and its something that you might want to look at. If it contained the data rows, that can sometimes be very large. 

Points 2 - Yes I probably prefer this style. I did consider returning a success value previously, but if we are happy flanking try/except on sim/gen work then we probably wont be doing that.

jlnav:
> If I'm reading this right, H_sel would contain the actual data. I guess its a view but it would need to go with a single send, which needs revisiting comms. So I would keep work as metadata, and its something that you might want to look at. If it contained the data rows, that can sometimes be very large.

Fair enough!

Another question: Following an ``InsufficientFreeResources`` exception, why ``break`` instead of ``continue``?

shuds13:
> Another question: Following an `InsufficientFreeResources` exception, why `break` instead of `continue`? Are other work-units for either the sim or gen also unlikely to succeed, even if they (for sims) potentially request smaller resource sets?

They will be requesting the same resource sets, because it will keep trying to schedule the same work. Generally each loop is dealing with either sims or gens (or only doing a gen once all work is done). Let me know if you spot an exception.

I have an alternative for preemptive algorithm, but I've not pushed it. In that case there is an inner loop that will go though the available work units until either enough resources is found or has gone through all work generated.

jmlarson1:
This is being completed by #664

PR #662, opened 2021-09-16
Small rset_team refactor within AllocSupport

jlnav:
Makes `rset_team` assignment to `libE_info` in `sim_work` and `gen_work` an additional function. Perhaps not necessary, but reduces a handful of redundant lines.


shuds13 (commented):
I think I prefer num_rsets_req as being more specific than parsed_resources, otherwise I'm good with this.

jmlarson1:
I like `num_rsets_req`

PR #665, opened 2021-09-21
tentatively rename allow_oversubscribe, document

jlnav:


shuds13 (commented):
A grep shows couple of mentions of allow_oversubscribe still in docs

PR #669, opened 2021-09-24
Persistent Support into Persistent Sim Example

jlnav:
Looks like develop needs to be pulled into the target branch, then the new class functions need updating based on progress in `gen_support.py` Please let me know if I can help out with this in any way


shuds13 (commented):
Not sure where this branch was taken from, but unless you have an easier answer, I suggest that I remove the experimental branch, and I will add the persistent sim changes on top of this. We can pull this into develop first perhaps?

jlnav:
> Not sure where this branch was taken from, but unless you have an easier answer, I suggest that I remove the experimental branch, and I will add the persistent sim changes on top of this. We can pull this into develop first perhaps?

This was taken off the experimental branch, but I also merged develop in. Sounds good, I'll close and re-open merging into develop.

PR #670, opened 2021-09-27
Adds `persis_in`, removes `alloc_specs['in']`, various small refactors

jlnav:


libensemble/alloc_funcs/persistent_aposmm_alloc.py
```diff
@@ -33,7 +33,7 @@ def persistent_aposmm_alloc(W, H, sim_specs, gen_specs, alloc_specs, persis_info
         assert support.all_given_back(H), "Initial points in H have never been given back to gen."
 
         # SH TODO: gen_specs['persis_in']
-        persis_info['fields_to_give_back'] = ['f'] + [n[0] for n in gen_specs['out']]
+        persis_info['fields_to_give_back'] = gen_specs['persis_in']
 
         if 'grad' in [n[0] for n in sim_specs['out']]:
```
shuds13, line 38: @jmlarson1 @jlnav  Do we want to construct fields to give back like this - or expect the user to do it in calling script when creating gen_specs['persis_in'] ? An alloc could support a default, if the user does not set persis_in maybe, but seems to me potentially confusing to mix user supplied and add to it in alloc.

jlnav: Yeah, I'd lean towards expecting the user to define their fields in their calling script. I think I was more explicit with all the other cases except the new functions from the student.

Any major objections to refactoring the above conditional fields to be in their corresponding calling scripts instead?

jmlarson1: I also think that we should expect the user to define their fields in their calling script.

jmlarson1 (approved):

libensemble/alloc_funcs/persistent_aposmm_alloc.py
shuds13: Sounds good to me.

shuds13 (approved):

PR #671, opened 2021-09-27
Feature/persis worker support class

jlnav:
Pulling into develop instead


shuds13 (commented):
@jmlarson1 @jlnav I think I like this. One question is
whether we should model names on MPI?
should it be receive() or just recv() as in MPI?
like wise send_and receive - there is an mpi sendrecv()

Or do you prefer written out like this.

jmlarson1:
I thought we wanted to move away from MPI-exclusive terminology. We did with `ranks` and `processes` and some other items as well.

shuds13:
> I thought we wanted to move away from MPI-exclusive terminology. We did with `ranks` and `processes` and some other items as well.

the `procs_per_node` was really more to be consistent with the `num_procs` - its also used in MPICH as `-ppn` (processes/ors per node). 

In any case, I dont think we need to model on MPI here, but `recv` is also used in python multiprocessing so seems kind of standard. I dont feel that strongly about this, but I'm leaning towards `recv`

jlnav:
> In any case, I dont think we need to model on MPI here, but `recv` is also used in python multiprocessing so seems kind of standard. I dont feel that strongly about this, but I'm leaning towards `recv`

Looked at the multiprocessing API, yeah I'd be fine with `recv` too. I'm also okay with `send_recv`

shuds13 (approved):
I approve on the change to recv.

PR #672, opened 2021-09-28
Pass "libE_info" to allocation functions

jlnav:
Initial attempt to pass "libE_info" to allocation functions, containing selected summations of points from H, exit criteria, etc.


jmlarson1:
Were we going to have a different PR where those fields were used? 

Is it possible to make them optional, or should we make everyone change their alloc_f calling syntax? (I'm fine either way)

jlnav:
> Were we going to have a different PR where those fields were used? 

We could. The most important value now passed in is `exit_criteria`, right?

> Is it possible to make them optional, or should we make everyone change their alloc_f calling syntax? (I'm fine either way)

It's looking like everyone may have to change their `alloc_f` code anyway for `AllocSupport`, which is much less optional. I'm leaning towards everyone changing their `alloc_f` calling syntax; actually using the values passed in is optional of course.

shuds13:
The condition in work_giving_term_test() is 

```python
if ('sim_max' in self.exit_criteria
              and self.hist.given_count >= self.exit_criteria['sim_max'] + self.hist.offset)
```

As this is slightly non-trivial we could provide this as a boolean like `sim_max_given` and then in alloc would just be

```python
if sim_max_given:
  return ....
```

Any preferences? @jlnav @jmlarson1

jlnav:
> So we could use this to tell all allocs to return if given_count > sim_max. This could replace the term_test in the middle of the manager run loop.

Maybe I misunderstand, but it already looks like the user can return a non-zero flag whenever and break the main loop as effectively as `term_test()` would when tripped. I was thinking we provide users the stats to conditionally return a flag, but not replace the term_test in case they're not interested in writing that code. Thoughts?

Also yeah, the new offsets can go in the alloc's libE_info.

shuds13:
The goal is not to break out of the loop. It is to stop creating more work. Currently the test in line 592 in the current branch cancels work packets that are created after sim_max is given, but note it is only the Work loop it breaks from. This means that work keeps getting packed up and cancelled after sim_max is given.

jmlarson1:
@shuds13 I am okay with passing in of a boolean `sim_max_given` or `elapsed_time_hit` or other tests. I think those booleans are all that would be needed (as opposed to the raw values themselves). 

If a user wanted the raw values, they could compute them from H I believe.

shuds13:
> If a user wanted the raw values, they could compute them from H I believe.

They could but its filter a whole column of H. So I was thinking of performance.

jmlarson1:
I think we agree. Giving the boolean `sim_max_given` is saving the user from filtering an entire column of `H` which is good for most.

shuds13:
Yes, I see the point. We agree on passing the booleans.

I was thinking about the given / returned / given_back counts which could also be useful in other places. I'm assuming we agree about that also.

jmlarson1:
Yup. Any of those already-computed-in-the-manager quantities would be great to give to the `alloc_f`in libE_info

libensemble/manager.py
```diff
@@ -534,6 +521,22 @@ def _final_receive_and_kill(self, persis_info):
 
     # --- Main loop
 
+    def _get_alloc_libE_info(self):
+        "Selected statistics useful for alloc_f"
+
+        libE_info = {'exit_criteria': self.exit_criteria,
+                     'elapsed_time': self.elapsed(),
+                     'manager_kill_canceled_sims': self.kill_canceled_sims,
```
jmlarson1, line 529: Are you going to be modifying this PR to show where each of these can be used?

jlnav: I am going to describe each of these fields in the docs and suggestions on how they could be used in an allocation function, but were you thinking of allocation function examples?

shuds13 (commented):
Suggest changing work_given_tripped to sim_max_given.

libensemble/manager.py
shuds13: I dont think we had a specific case for these, but in give_sim_work_first I think we were going to do Jeff's new cancel block after sim_max_given. 

We also talked about bringing the condition "if any(self.W['active'] == 0)" inside here - perhaps git could be made a boolean?

shuds13: Okay - it will get points_to_evaluate so may as well have a quick check at the top.

jlnav:
Lots of additional docs changes have been completed, and the idle worker check beginning every alloc is now a bool computed on the manager side and passed in libE_info (although of course users are still free to compute whatever they like out of H and W in their allocs)

shuds13:
So I think just need to move the cancel block into the IF at the top in give_sim_work_first if that works. Then if ttests pass I approve this.

shuds13:
> So I think just need to move the cancel block into the IF at the top in give_sim_work_first if that works. Then if ttests pass I approve this.

I actually mean above the first block - I've implemented this.

PR #673, opened 2021-09-29
Additional offsets for term_test comparision 

jlnav:
Probably need to made a copy of `test_evaluate_existing_sample.py` with an H0 with different mixes of given, returned, etc. points.


jlnav:
I'd prefer to pull this in first so the offsets can be incorporated into statistics inside the alloc's libE_info

jmlarson1 (approved):

PR #674, opened 2021-09-30
Add persistent sim example.

shuds13:
Ensures libE_info used (including H_rows) updated for each sim in loop (and returned libE_info updated in PersistentSupport)
Split use_persis_return to use_persis_return_sim and use_persis_return_gen

Does not:
  include ability to send work request with persistent stop.
  fix up returns with correct libE_info on persistent stop/return from persistent sim function


shuds13:
@jmlarson1 A problem - when we send a PERSIS_STOP tag along with data, it does not get marked as given because in manger, it uses the tag:

https://github.com/Libensemble/libensemble/blob/39f69adca6fb38f1873c07e2652c66e58a3fece9/libensemble/manager.py#L359-L362

This is how we distinguish whether its a sim or gen at this point.

PR #678, opened 2021-10-02
Docs/cleaning up all tests

jmlarson1:
Our regression tests doc strings and other formatting choices were inconsistent. 

I'd like the doc strings to tell more about what libEnsemble capability is being tested.


shuds13:
@jmlarson1 If your cleaning doc strings, I should point out this line.

`The number of concurrent evaluations of the objective function will be 4-1=3`

is generally not true for any persistent gen tests. In most cases its 4-2=2

jmlarson1:
> @jmlarson1 If your cleaning doc strings, I should point out this line.
> 
> `The number of concurrent evaluations of the objective function will be 4-1=3`
> 
> is generally not true for any persistent gen tests. In most cases its 4-2=2

Very good catch. I will adjust. Feel free to recommend more changes!

PR #681, opened 2021-10-02
Fix up for persis sim return

shuds13:
Allows a return from persistent sim to be included in the correct H_row.

This uses storing in the comm which worker already has access to. We already do the same thing with the push_to_buffer().

It seems to me this is a possible solution. Unless we subprocess the user functions and add a worker/user function communicatoin layer, then the comms object is a point of communication.


jlnav (approved):
Approved once unit test is fixed

PR #685, opened 2021-10-05
Do you like black more than yapf?

jmlarson1:
I like black here more than yapf. How about you?


jlnav:
``pip install dragonfly-op`` ?

In any case, I think I'm coming around to black. I especially like it's reformatting of H0s. I don't _think_ we need to install additional dependencies on the CI side, right? We could keep flake8 there and those who run black locally don't have to think about it anymore. Those who _don't_, can keep iterating with flake8 no problem.

jmlarson1:
1. I just trying pushing a direct-to-`develop` update to `.github/workflows/libE-ci.yml` with `pip install dragonfly`
2. I am not yet advocating for using `black` everywhere in the repo, nor having it be a CI dependency. Just thought it looked nice in the regression tests. I might try it in the sim/gen/alloc scripts as well.

PR #696, opened 2021-10-08
Docs/update readme and examples

shuds13:
I want to keep a very simple README in examples. I've moved tutorials in to subdirs. @jlnav can you double check I've not broken any links or anything.


setup.py
```diff
@@ -55,7 +55,7 @@ def run_tests(self):
 
     package_data={'libensemble.sim_funcs.branin': ['known_minima_and_func_values']},
 
-    install_requires=['numpy', 'psutil', 'pyyaml'],
+    install_requires=['numpy', 'psutil],
```
jlnav, line 58: missing tick after psutil

PR #699, opened 2021-10-14
Testing/fix executorhworld

jlnav:
Sorry this took so long to pull in - hopefully since I wasn't able to reproduce it locally these changes only reflect the slowness of CI and, if the release had happened earlier, the release package wouldn't miss out on any important changes


shuds13 (approved):
Looks like all tests passing and I've checked (non-yaml) tests work without yaml installed.

PR #712, opened 2021-11-08
Experimental/funcx

jlnav:
Addresses #713 

Initial attempts to draft an optional funcX interface so workers can easily launch _non-persistent_ user functions on remote resources:

https://funcx.org/

Some notes:

- Security is taken care of by funcX. Users must authenticate with Globus when initializing an endpoint.
- This user-prompted security makes CI testing currently-impossible?: https://github.com/funcx-faas/funcX/issues/619
- funcX endpoints have a limit of 20 function calls every 10 seconds unless batching is used.
- Endpoints can only be instantiated on Linux. If using clusters, endpoints can (and should) be configured by the user to launch their functions to compute nodes. Many examples, including for Theta, are available on the funcX docs.
- This (currently) isn't an alternative to MPI or local comms, or any Executor. This currently only allows user functions to run on a different resource than libEnsemble's processes.

To try this out locally:

1.  ``pip install funcx funcx-endpoint``
2. ``funcx-endpoint configure my-endpoint``
3. (You may be asked to authenticate with Globus. Do so using the funxc-endpoint generated URL)
4. ``funcx-endpoint start my-endpoint``. A message will confirm startup and print the endpoint's uuid.
5.  Set ``sim_specs['funcx_endpoint']`` to this uuid.
6. Run like normal: ``python test_funcx.py --comms local --nworkers 4``

Documentation and/or other improvements coming soon?

TODO:

- [x] Integration with yaml interface
- [x] Test use with executor
~~- [ ] Test cancellations~~ Comm is necessary for manager.poll - so this doesn't work. function cancellation is a work-in-progress with the funcX team: https://github.com/funcx-faas/funcX/pull/606
- [x] Documentation


jmlarson1 (approved):
I'd remove the changes in the alloc_fs and make it so the manager doesn't error when receiving a `None` persis_info

jmlarson1:
Thoughts on loss of coverage in openmpi runners?

jlnav:
> Thoughts on loss of coverage in openmpi runners?

still need to fix it by allowing the executor_hworld tests to run with local comms

jlnav:
I've completely rewritten the funcX test to be a scaling test. Try this out on Theta:

```
pip install funcx-endpoint
funcx-endpoint configure forces
```

Configure ``~/.funcx/forces/config.py`` to have the following:

```
from parsl.addresses import address_by_hostname
from parsl.launchers import AprunLauncher, SimpleLauncher
from parsl.providers import CobaltProvider

from funcx_endpoint.endpoint.utils.config import Config
from funcx_endpoint.executors import HighThroughputExecutor

# fmt: off

# PLEASE UPDATE user_opts BEFORE USE
user_opts = {
    'theta': {
        'worker_init': 'source ~/startup.sh',
        'scheduler_options': '',
        # Specify the account/allocation to which jobs should be charged
        'account': 'CSC250STMS07'
    }
}

config = Config(
    executors=[
        HighThroughputExecutor(
            max_workers_per_node=1,
            worker_debug=False,
            address=address_by_hostname(),
            provider=CobaltProvider(
                queue='debug-flat-quad',
                account=user_opts['theta']['account'],
                #launcher=AprunLauncher(),
                launcher=SimpleLauncher(),

                # string to prepend to #COBALT blocks in the submit
                # script to the scheduler eg: '#COBALT -t 50'
                scheduler_options=user_opts['theta']['scheduler_options'],

                # Command to be run before starting a worker, such as:
                # 'module load Anaconda; source activate funcx_env'.
                worker_init=user_opts['theta']['worker_init'],

                # Scale between 0-1 blocks with 2 nodes per block
                nodes_per_block=1,
                init_blocks=0,
                min_blocks=0,
                max_blocks=1,

                # Hold blocks for 30 minutes
                walltime='00:10:00'
            ),
        )
    ],
)
```

Then run:

``funcx-endpoint start forces``

On your local machine (or any machine with internet access), replace ``sim_specs['funcx_endpoint']`` with the printed uuid. See ``funcx_forces.yaml``. Adjust the other paths in that file as necessary.

Then on your machine where you'll run libEnsemble, run the calling script as usual. As long as the local ``exit_criteria`` have not been met, funcX will dynamically submit allocations via cobalt to the specified queue in the above config, then run the sim_f.

Using ``launcher=SimpleLauncher(),`` in the above config means the simulator function will be run on the MOM nodes as usual. Replace it with ``launcher=AprunLauncher(),`` to experiment with submitting the sim_f directly to the allocated compute nodes instead!

libensemble/manager.py
```diff
@@ -433,8 +433,9 @@ def _update_state_on_worker_msg(self, persis_info, D_recv, w):
             else:
                 self._freeup_resources(w)
 
-        if 'persis_info' in D_recv and len(D_recv['persis_info']):
-            persis_info[w].update(D_recv['persis_info'])
+        if D_recv.get('persis_info') is not None:
```
shuds13, line 436: Remind me - why not just if D_recv.get('persis_info') instead of 436/437

shuds13:
Wondering if we can avoid some of the duplication of forces stuff. Also looks like there is a binary there (forces.x).

libensemble/manager.py
jlnav: That's probably valid. The original work was to address how persis_info is handled by the manager if the user didn't specify one but the alloc tried packing it up anyway.

jlnav:
> Wondering if we can avoid some of the duplication of forces stuff. Also looks like there is a binary there (forces.x).

addressed

shuds13 (approved):

PR #715, opened 2021-11-16
Testing workflow experimental refactors

jlnav:
1. what happens if only complete.yml tells coveralls that all jobs are complete, since it's the last workflow to finish?


jlnav:
hallelujah coverage!

jmlarson1 (approved):

PR #717, opened 2021-11-19
Add Ax Multi task algorithm

RemiLehe:
This adds an example with the Ax multi-task optimizer.
Many thanks to @AngelFP for implementing the hooks between `Ax` and `libEnsemble`.


shuds13:
@jmlarson1 @jlnav This is currently in scaling tests, so not tested on CI. I have run locally. I'm wondering whether we should put this in our regression tests currently. That way it would get tested on CI. Though we may move to our new examples repo eventually. It has a few dependencies for the gen, but they seemed to be quite quick to install. The sim is simple in this example.

shuds13:
The only downside I think is a few dependencies for ax. https://ax.dev/docs/installation.html.
This was quite quick to install for me locally/

shuds13:
I suggest we merge into develop, with those tests failing. Then we can make a branch to do the updates to 3.6 (and 3.10) testing.

shuds13 (approved):

jmlarson1 (approved):

Issue #610, opened 2021-05-06
Implement Python's Concurrent Futures interface into Executor


jlnav:
Will contribute towards compatibility with other Exaworks packages, among other benefits

jlnav:
Closed by #719

PR #719, opened 2021-12-10
Concurrent futures interface for Executor

jlnav:
Addresses #610 

Very early approaches. Non-breaking changes, all new features should be alternatives or wrappers for existing functions.


libensemble/tests/unit_tests/test_executor.py
```diff
@@ -688,6 +688,37 @@ def test_serial_startup_times():
     assert 0 < startup_time < 1, "Start up time for python program took " + str(startup_time)
 
 
+def test_futures_interface():
+    print("\nTest: {}\n".format(sys._getframe().f_code.co_name))
+    setup_executor()
+    exctr = Executor.executor
```
shuds13, line 694: Do you need this line as well as 697? (same for 711)

libensemble/executors/executor.py
```diff
@@ -254,6 +254,25 @@ def wait(self, timeout=None):
...
+        self.wait(timeout=timeout)
+        return self.state
+
+    def running(self):
+        """ Return ```True`` if task is currently running."""
+        return self.state == 'RUNNING'
+
+    def done(self):
+        """ Return ```True`` if task is finished or successfully cancelled."""
+        return self.state in ['FINISHED', 'USER_KILLED']
```
shuds13, line 274: I'm wondering if our state of FAILED should also return True? 
I may have made this a bit confusing. I think self.finished == True includes FAILED, but a self.state of FINISHED is different to a state of FAILED. I think wrapping self.finished may be what we want.

shuds13:
So we should be clear this so far does  not support everything in https://docs.python.org/3/library/concurrent.futures.html but it does support the most common future functions.

jlnav:
I'm personally happy with what's currently been added. If it's worth adding ``map()`` or other functions we can revisit that in another PR. Thoughts appreciated of course.

shuds13 (approved):

PR #720, opened 2021-12-13
Experimental/multi task regtest

shuds13:
Branch to check multi-task branch working as regression test.


shuds13:
Need to add missing dependencies to setup.py 'extras'.

PR #722, opened 2021-12-14
Docs / JLN potential docs updates

jlnav:
Branched off Steve's docs branch.

At this point, primarily attempts a modernized rewrite of the README. I'm still finding ways to simplify the language but emphasize what needs to be said.


shuds13:
Note that I took  #723 off here so will pull into this branch.

README.rst
```diff
@@ -25,35 +25,41 @@
...
+
+• **Extreme scaling**, from laptops to supercomputers.
+• **Resilience/fault tolerance**. libEnsemble can restart incomplete tasks or entire ensembles.
+• **Monitoring of applications**, including recovering and dynamically reassigning resources.
+• **Portability and flexibility**. Run identical libEnsemble scripts across machines and facilities.
+• **Exploitation of persistent data/control flow**. libEnsemble can pass data between running functions.
+• **Relatively low start-up cost**. Most single-machine deployments don't require additional services.
+
+libEnsemble employs a manager/worker scheme that runs on MPI, multiprocessing,
+or TCP. Workers control and monitor any level of work using the aforementioned
```
jmlarson1, line 56: What are the "aforementioned functions"?

jmlarson1 (approved):

README.rst
jlnav: sims and gens. I'll make that more clear

jlnav:
Just noticed that "and across" was removed from right before "leadership-class machines and facilities" in the README, that is technically possible now with funcX integration, and will be further possible with the Balsam executor updated in the future. Any thoughts on or not emphasizing "across machines" at this time?

jmlarson1:
The "and across" deserves clarification (if we keep it). Do we have (or want) an example showing where we can run on all of the LCF systems at the same time?

jlnav:
> 

I specify funcx as an optional feature below, but maybe its worth waiting until Balsam has also been incorporated and more funcX configurations have been tested before we state resolutely that "across" is well supported.

As for an example, I wouldn't want to try "all of them at the same time", but Theta+Summit (or any combination of LCF systems) may be possible and interesting. A ddmd or warpx example could be modified to run sims and gens on separate systems

README.rst
shuds13: Maybe add: Cancellation/killing of simulations.

README.rst
```diff
@@ -206,10 +203,7 @@ If running in local mode, which uses Python's multiprocessing module, the
 ``local`` comms option and the number of workers must be specified. The script
 can then be run as a regular Python script::
 
-    python myscript.py
-
-These options may be specified via the command line by using the ``parse_args()``
-convenience function within libEnsemble's ``tools`` module.
+    python myscript.py --comms local --nworkers N
```
shuds13, line 206: Not sure we should remove this, as the added line assumes ``parse_args``. I do think that before it was a bit insufficient also, and maybe could do with a link to example of setting libE_specs for local comms. But as this is just an overview, maybe there is a concise way to express.

README.rst
```diff
@@ -218,14 +212,16 @@ Resources
 
 **Support:**
 
-- The best way to receive support is to email questions to ``libEnsemble@lists.mcs.anl.gov``.
```
shuds13, line 221: As said, I would rather have email/slack first. Along with the improved access we talked about.

shuds13: At first we could say - email us for a slack invite. Then maybe update to the form method.

docs/function_guides/simulator.rst
```diff
@@ -48,3 +48,13 @@ and monitor applications. An excellent overview is already available
...
+
+Persistent Simulators
+---------------------
+
+Although comparatively uncommon, simulator functions can also be instantiated and written
+in a persistent fashion. See the :ref:`here<persistent-gens>` for a general API overview
+of writing persistent generators, since the interface is largely identical. The only
+differences are to include ``EVAL_SIM_TAG`` when instantiating a ``PersistentSupport``
+class instance, and to return ``FINISHED_PERSISTENT_SIM_TAG`` when the simulator
+function returns.
```
shuds13, line 60: Perhaps a link to the persistent sim example
eg. https://github.com/Libensemble/libensemble/blob/main/libensemble/tests/regression_tests/test_persistent_sim_uniform_sampling.py

README.rst
shuds13: Perhaps use the block from running libE in manual. https://libensemble.readthedocs.io/en/docs-jln_docs_updates/running_libE.html

docs/running_libE.rst
```diff
@@ -66,8 +66,7 @@ where ``N`` is the number of workers. This will launch one manager and
 
 libEnsemble will run on one node in this scenario. It is only suitable for running
 in :doc:`central mode<platforms/platforms_index>` on multi-node systems. It can also
-be used on stand-alone systems. *Technically, you could run without dedicated_mode set,
```
shuds13, line 69: Why remove? 
I think the central v distributed was a bit of bad distinction. Some users, including WarpX/FBPIC users run libEnsmeble on one node and share with launched applications, while other nodes will just run launched applications. The first node does not conflict, because their WarpX runs are limited by number of GPUs (with one cpu per gpu), so there is plenty to run libEnsemble on that node. This means that my original line that its only suitable for 'central mode', given our definition, is not true. So I think we could re-word better. 
Dedicated v shared for libEnsemble might ultimately be better words.

README.rst
```diff
@@ -25,35 +25,43 @@
...
-communication media (including MPI, multiprocessing, and TCP); interfacing with
-user-provided executables is also supported. Each worker can
-control and monitor any level of work, from small subnode tasks to huge
-many-node simulations. An executor interface is provided to ensure that scripts
-are portable, resilient, and flexible; it also enables automatic detection of
-the nodes and cores available to the user, and can dynamically assign resources
-to workers.
+libEnsemble is a Python_ toolkit for coordinating asynchronous and dynamic ensembles of calculations.
+
+libEnsemble workflows can run on and across laptops, leadership-class machines and facilities,
```
shuds13, line 30: Yeah, I think the across is not that clear. I think I would say
"libEnsemble workflows can run on laptops, clusters, and leadership-class machines,"

or to generalize a bit more could say
"libEnsemble workflows can run on diverse platforms including laptops, clusters, and leadership-class machines,"

shuds13: If you want to keep 'across', thats okay.

shuds13: libEnsemble workflows can run on or across  laptops, clusters, and leadership-class machines

jmlarson1 (approved):

shuds13 (approved):

PR #723, opened 2021-12-21
Update history docs inc. workflow example

shuds13:
Updating history array docs and adding example workflow with diagrams.

Branched of here as already a docs branch. 
I recommend a squash merge as image files were added and then replaced.

TODO: 

- [x] Need to redo images as unitialized points should have sim_id of -1.
- [x] Deal with comments inline (history_array.rst).
- [x] Evaluate image data type (is svg better).


shuds13:
@jmlarson1 @jlnav 
Please take a look at https://libensemble.readthedocs.io/en/docs-history_update/data_structures/history_array.html

Questions:

1. Language: Do we call the inputs/outputs of sim/gen history arrays (e.g. this data structure) - or does this term just refer to the manager;s history  array. This could affect my wording a bit.

2. Should we separate out protected fields in the listing. Or otherwise re-order?

docs/data_structures/history_array.rst
```diff
@@ -7,25 +7,76 @@ history array
...
+Each entry contains multiple named fields, which can contain different data types,
+including scalar and arrays.
+
+The manager maintains the global history array for the ensemble. Each entry stores
+data generated by the ``gen_f`` and then the resultant output from the ``sim_f``,
+in addition to some reserved fields.
+
+When the history array is initialized, it creates fields in ``H`` for each entry of
+``gen_specs['out']`` and ``sim_specs['out']``, provided by the user. For a simple
+example these fields might be::
```
jmlarson1, line 24: Can we link to these reserved names?

docs/data_structures/history_array.rst
```diff
@@ -7,25 +7,76 @@ history array
...
-final history array returned by libEnsemble will include the following
-fields, almost all of which are protected from the user. These fields are
-populated internally:
+store information about each point (ensemble member) processed in the ensemble.
+Each entry contains multiple named fields, which can contain different data types,
+including scalar and arrays.
+
+The manager maintains the global history array for the ensemble. Each entry stores
+data generated by the ``gen_f`` and then the resultant output from the ``sim_f``,
+in addition to some reserved fields.
```
jmlarson1, line 20: Can we link to the reserved fields here?

docs/data_structures/history_array.rst
```diff
@@ -7,25 +7,76 @@ history array
...
+(since the epoch) corresponding to each state, and when the point was generated.
+Other protected fields include the worker IDs on which points were generated or
+evaluated.
+
+The user fields and the reserved fields together make up the final history array
+returned by libEnsemble.
+
+The full list of these reserved fields is given below.
+
+.. TODO separate list by protected field or field a user can set]
```
jmlarson1, line 69: In this PR?

docs/data_structures/history_array.rst
shuds13: - [x] Added a referenced to reserved fields section.

Also made gen_f/sim_f references - though not sure whether should go to API (as I did) or the function writing guides eg. https://libensemble.readthedocs.io/en/docs-history_update/function_guides/simulator.html

docs/data_structures/history_array.rst
shuds13: I have done this - what do you think.

jmlarson1: I think the break-down is right, but I wouldn't say that `sim_id` is "modifiable", because it can either be set by the user or by libE. (But it must also be in numerical order, so it's quite constrained.)

docs/data_structures/history_array.rst
jmlarson1: I think it is good.

docs/data_structures/history_array.rst
shuds13: I wasn't sure what word to use here.

shuds13:
@jmlarson1 @jlnav 

Should we move history up the TOC tree here - may even at the top. https://libensemble.readthedocs.io/en/docs-history_update/data_structures/data_structures.html
or possibly after sim/gen/alloc

jmlarson1:
I think it can be at the top, yes.

jlnav (approved):

docs/data_structures/history_array.rst
shuds13: I've removed so only 'Proected fields' sub-heading.

Issue #631, opened 2021-06-17
Ensure libEnsemble supports most recent Balsam versions


jlnav:
Update libEnsemble so it supports most recent Alpha (0.7.0a) and Release (0.5.0) versions of Balsam. Document examples and guides for using libEnsemble with most recent versions, including possible use-cases and routines for cross-system task submissions.

- [x] 0.5.0

- [x] 0.7.0a

PR #729, opened 2022-02-08
Feature/balsam7

jlnav:
Addresses #631 

Opening this to share. Still a work-in-progress.

This is a complete rewrite of the Balsam MPI Executor (currently a separate file) to incorporate the changes and new capabilities of Balsam2. The biggest new capability is workers can submit tasks from any machine, to any machine. The Balsam service, via Balsam sites, takes care of reserving resources and launching the tasks.

Regarding the interface, thankfully, the ``executor.submit()`` method is _pretty close_ to what we're used to with our traditional ``MPIExecutor``, so app-launching user-functions shouldn't change much. Within calling scripts, however, there's two major changes:

1. Users must register Balsam ``ApplicationDefinition`` instances to the executor instead of just paths to executables. This is because Balsam apps are now inherently and inseparably tied to their class name, their site, and other class attributes. ``ApplicationDefinition``s are literal-parsed, serialized, and sent to the Balsam service on class instantiation. An example in a calling script may resemble:
```
from balsam.api import ApplicationDefinition

class RemoteForces(ApplicationDefinition):
    site = 'thetaGPU'
    command_template = '/path/to/forces.x {{sim_particles}} {{sim_timesteps}} {{seed}} {{kill_rate}}'

exctr = NewBalsamMPIExecutor()
exctr.register_app(RemoteForces, app_name='forces')
```

2. Users must submit batch resource requests corresponding to sites where apps should run. This can be done with the new ``submit_allocation()`` function _in a calling script_, unless each worker wants to submit to a separate allocation or machine for some reason. An example may resemble:

```
exctr.submit_allocation(site_id=999, num_nodes=256, wall_time_min=300,
                        queue='default', project='my-project-code')
```

I _don't think_ libE itself necessarily needs to be launched with Balsam anymore, but I need to test this. Take a look at the ``scaling_tests/balsam_forces`` example.

TODO:

- [x] General refactoring and cleanup
- [x] Documentation
- [x] Address feedback
- [x] Test how well task.poll() is actually working
- [x] Update balsam_forces README
- [x] Executor.revoke_allocation()

Longer term TODO (I think these should go in another PR, probably for 0.9.1):

- Possibly also register _user functions_ with Balsam. Balsam ``ApplicationDefinitions`` can _either_ have a ``command_template`` field to launch an executable like we're familiar with, _or_ a ``run`` field to launch a Python function.
- Update templater for new Balsam tests


libensemble/tests/scaling_tests/balsam_forces/run_libe_forces_balsam.py
```diff
@@ -0,0 +1,44 @@
...
+forces.gen_specs["user"].update(
+    {
+        "lb": np.array([0]),
+        "ub": np.array([32767]),
+    }
+)
+
+forces.persis_info.add_random_streams()
+
+if forces.is_manager():
```
shuds13, line 21: I dont think this should be a function

libensemble/tests/scaling_tests/balsam_forces/readme.md
```diff
@@ -0,0 +1,61 @@
+## Running test run_libe_forces_balsam.py
+
```
shuds13, line 2: We might want to think about how we structure the variants of the forces example, and maybe have one place that describes the test, and its purpose, in a generic way.

libensemble/tests/scaling_tests/balsam_forces/readme.md
```diff
@@ -0,0 +1,61 @@
...
+See below.
+
+### Running with libEnsemble.
+
+On the remote machine:
+
+    git clone https://github.com/argonne-lcf/balsam.git
+    cd balsam; pip install -e .
+    cd ..; balsam site init ./my-site
+
```
shuds13, line 34: I would add `balsam login` before `balsam site init ./my-site`. As that line needs to be run afterwards.

libensemble/tests/scaling_tests/balsam_forces/readme.md
```diff
@@ -0,0 +1,61 @@
...
+Total energy is appended to the forces.stat file.
+
+To run forces as a standalone executable on `N` procs:
+
+    mpirun -np N ./forces.x <NUM_PARTICLES> <NUM_TIMESTEPS> <SEED>
+
+This application will need to be compiled on the remote machine where the sim_f will run.
+See below.
+
+### Running with libEnsemble.
```
shuds13, line 27: We want to say up front, that you currently need ALCF login credentials to use this service, regardless of what remote machine you are using. 

Then we could say here is an example running libEnsemble on your local computer/laptop and having `forces.x` runs on the remote machine. Making clear which instructions are for laptop and which for remote machine. E.g. Any changes to calling script/sim function are on laptop. But forces.x must be compiled on remote machine.

shuds13: You may want to tell the user to set up their own conda (or other virtual) environment. I first tried the pip install of balsam without that and got a warning could not access /soft/datascience/conda/miniconda3/latest/lib/python3.7/site-packages/.

Also they will need balsam installed on their local machine.

libensemble/tests/scaling_tests/balsam_forces/readme.md
```diff
@@ -0,0 +1,61 @@
...
+On the remote machine:
+
+    git clone https://github.com/argonne-lcf/balsam.git
+    cd balsam; pip install -e .
+    cd ..; balsam site init ./my-site
+
+You may be asked to login and authenticate with the Balsam service. Do so with
+your ALCF credentials.
+
+Configure the `RemoteForces` class in the `run_libe_forces_balsam.py` calling
```
shuds13, line 38: On local machine.

libensemble/tests/scaling_tests/forces/run_libe_forces.py
```diff
@@ -38,12 +38,15 @@
 
 # Create executor and register sim to it.
 if USE_BALSAM:
-    from libensemble.executors.balsam_executor import BalsamMPIExecutor
```
shuds13, line 41: Are the changes here (under regular forces directory) for running balsam on the local machine?

shuds13:
Some notes:

This has worked for me without the reading output file and killing tasks.

Reading output file and killing. The original Balsam has functions to do this which are used in the `balsam_executor.py`. I wonder if this can be done remotely in this balsam - albeit with a larger overhead, which may change how often you want to do these interactions. 

In the Executor initialization we inherit from MPIExecutor code to determine things like MPI runner. This is currently not passed through to Balsam so will be ignored (also in existing balsam executor). In this case, any detected stuff (MPI runners or resources) is going to be bad (even if it was passed through to Balsam), as it is looking on the local machine. So if we did want to do resource detection it would happen on the remote system and would prevent oversubscription to Balsam. Currently, we should probably just allow Balsam to do everything - maybe disable resource detection. That said, it is possible a user may want to tell you (via custom_info) what MPI runner they want to use (e.g. srun or mpirun which are both present on some systems - with different options). So we should see if Balsam accepts such requests, though may not be in scope for this PR.

libensemble/tests/scaling_tests/balsam_forces/readme.md
shuds13: And then you need to switch to the site directory and run:

`balsam site start`

jlnav:
Actually, I first need to test that we can use Balsam on Theta like we have _traditionally_, where libEnsemble runs on a compute node and submits additional Balsam jobs to the other nodes

docs/executor/balsam_executor.rst
```diff
@@ -1,17 +1,35 @@
...
+.. autoclass:: NewBalsamExecutor
+  :show-inheritance:
+  :members: __init__, register_app, submit_allocation, revoke_allocation, submit
+
+.. autoclass:: NewBalsamTask
+  :show-inheritance:
+  :member-order: bysource
+  :members: poll, wait, kill
+
+Balsam 1 MPI Executor
```
jmlarson1, line 19: Is there a reason "MPI" is here and not above?

libensemble/executors/new_balsam_executor.py
```diff
@@ -0,0 +1,505 @@
+"""
```
jmlarson1, line 1: Should this file be `balsam_2_executor.py` and rename the old one to balsam_1_executor.py`?

docs/executor/balsam_executor.rst
jlnav: Our older balsam executor implementation only ever used MPI I believe, but the new one can also do serial jobs. It also doesn't inherit from the mpi executor

libensemble/executors/new_balsam_executor.py
jlnav: I've done: the new one is ``balsam_executor.py`` and the old one is ``legacy_balsam_executor.py``

libensemble/executors/balsam_executor.py
```diff
@@ -130,32 +179,41 @@ def poll(self):
...
+        """Waits on completion of the task or raises ``TimeoutExpired``.
 
         Status attributes of task are updated on completion.
 
         Parameters
         ----------
 
-        timeout:
-            Time in seconds after which a TimeoutExpired exception is raised"""
+        timeout: int
```
jmlarson1, line 215: Only an `int`, not a `float`?

jmlarson1 (approved):

libensemble/executors/balsam_executor.py
jlnav: I think ``int`` matches the other executors' APIs.

Plus, less than a second is way too quick to wait on a task that's submitted to, and then scheduled by Balsam across the internet. More than one second might as well be whole seconds

jmlarson1: Is there a check or issue if a non-int value is given?

jmlarson1: Or are such quantities cast to ints?

libensemble/tests/scaling_tests/balsam_forces/readme.md
```diff
@@ -0,0 +1,61 @@
...
+script to match the Balsam site name and the path to your `forces.x` executable.
```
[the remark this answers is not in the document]
shuds13: I still think this is confusing. I guess you can call the directory "my-site" anything you want. I don't know why that does not come up as the default site name (Instead I get the name of the login node). I assume there is a one-to-one mapping between this directory and a site.

libensemble/tests/scaling_tests/balsam_forces/readme.md
shuds13: `RemoteForces` class is now in `submit_libe_forces_balsam.py` - though the site name is given in the calling script. Maybe could be a bit clearer.

shuds13: also path to forces.x

jlnav:
We discussed this Balsam routine transferring back or creating an ensemble directory that maps onto the remote Balsam data directory as much as possible. While evaluating this I realized we'd potentially need to support combinations of a couple things. I'm also recording this for my own reference:

1. Running libEnsemble local, submitting remote apps:

- Forces: Must transfer its results back to the local directory.
- libEnsemble: App not launched, but must maintain the ensemble directory.


2. Running everything remotely

- Forces: If running on same machine as libEnsemble, doesn't need to perform own transfers
- libEnsemble: Can transfer entire remote ensemble directory back.

- OR:

- Forces: _Can_ transfer its results back to the local directory.
- libEnsemble: No transfer needed, but _submission script_ must maintain ensemble directory (!)

Supporting all of these circumstances in balsam_forces is further complicated by:

1. Multiple different versions of RemoteForces and LibensembleApp will need to be defined, since ApplicationDefinitions' transfers fields are stored as-is on Balsam's end.
2. Each script that registers apps or produces Balsam Jobs will need logic to determine which type of Balsam app to import based on where it's running, then parameterize transfers
3. How could a submission script set up local worker directories for Forces apps to transfer .stat files back into?

**_I think at this time_**, I'd like to make an executive decision that the most interesting/relevant use-case is running libEnsemble on one's laptop and submitting apps to a remote system. I'd prefer a single well-working example for a PR this large.  In the future I can produce additional examples for running libEnsemble on the compute nodes with Balsam, then dealing with the various transfers shenanigans.

shuds13 (approved):

PR #742, opened 2022-02-18
[Hotfix] Try to fix tests on main with superlu version specified

jlnav:


shuds13:
I changed this to ready and approved, given https://github.com/Libensemble/libensemble/actions/runs/1865798233

PR #744, opened 2022-02-21
Check messages from inactive workers.

shuds13:
Without this, workers that are never active do not report initiation to logger.

Should we do this?


jlnav (approved):
If it works I'm happy

PR #746, opened 2022-02-26
Feature/match slots option

shuds13:
Add option for whether user wants to only use matching slots across multiple nodes.

If not required, then scheduling can be more efficient.

Update: We discussed moving scheduler_opts to libE_specs. I would like to do this as a separate PR if we are doing. It will actually mean a modification to all allocation functions to pass via libE_info to the alloc_support, as it is not currently passed through.


libensemble/resources/scheduler.py
```diff
@@ -31,91 +40,167 @@ def __init__(self, user_resources=None, sched_opts={}):
...
+            A dictionary of scheduler options.
+            Normally passed via ``alloc_specs['user']['scheduler_opts']``
+
 
+        The supported fields for sched_opts are::
 
-        **Scheduler Options**:
+            'split2fit' [boolan]:
+                Try to split resource sets across more nodes if space is not currently
+                available on the mininum node count required. Allows more efficient
```
jlnav, line 51: small typo

libensemble/resources/scheduler.py
```diff
@@ -31,91 +40,167 @@ def __init__(self, user_resources=None, sched_opts={}):
...
         sched_opts: dict, optional
-            A dictionary of scheduler options. Normally passed via alloc_specs['user']
+            A dictionary of scheduler options.
+            Normally passed via ``alloc_specs['user']['scheduler_opts']``
+
 
+        The supported fields for sched_opts are::
 
-        **Scheduler Options**:
+            'split2fit' [boolan]:
```
jlnav, line 49: boolean

libensemble/resources/scheduler.py
```diff
@@ -31,91 +40,167 @@ def __init__(self, user_resources=None, sched_opts={}):
...
-        """Schedule resource sets to a work item if possible
+        """Schedule resource sets to a work item if possible.
 
         If the resources required are less than one node, they will be
         allocated to the smallest available sufficient slot.
 
-        If the resources required are more than one node, then will
-        attempt to find an even split. If no even split is possible,
-        then enough full nodes will be allocated to cover the requirement.
+        If the resources required are more than one node, then the
```
jlnav, line 77: "the the"

jlnav (approved):

jmlarson1 (approved):

libensemble/sim_funcs/six_hump_camel.py
```diff
@@ -148,19 +152,23 @@ def six_hump_camel_CUDA_variable_resources(H, persis_info, sim_specs, libE_info)
...
+            'Worker {}: CUDA_VISIBLE_DEVICES={}  \tnodes {} ppn {}  slots {}'.format(
+                libE_info['workerID'], os.environ["CUDA_VISIBLE_DEVICES"], num_nodes, cores_per_node, slots
             ),
             flush=True,
         )
     else:
         # Unless use a matching sub-set, but usually you probably don't want this
-        print('Error: Cannot set CUDA_VISIBLE_DEVICES when uneven slots on nodes {}'.format(resources.slots))
+        print('Error: Cannot set CUDA_VISIBLE_DEVICES when unmatching slots on nodes {}'.format(slots))
+        raise SixHumpException("Cannot set CUDA_VISIBLE_DEVICES when unmatching slots on nodes.",
```
shuds13, line 169: Why not use an assert beforehand?

```python 
assert resources.matching_slots, "Error: Cannot set CUDA_VISIBLE_DEVICES when unmatching slots on nodes {}'.format(slots)"

```

shuds13: Done

PR #760, opened 2022-03-21
Feature/improved lib E field naming

jmlarson1:
These fields have many positives:
- They align with `sim_f` and `gen_f` and `sim_specs` and `gen_specs` and the like. 
- They tell a "story" about the entry in `H`.

1. The point is generated by `gen_worker` between `gen_start_time` and `gen_end_time`.
2. It is given to `sim_worker` and `sim_given` is marked true and `sim_given_time` is recorded
3. When `sim_worker` is done, `sim_completed` is marked true, and `sim_completed_time` is recorded.
4. If the `gen_worker` is informed about the point, then `gen_informed` is marked true and `gen_informed_time` is recorded. 

In the corner case where a point is requested again by a generator, we mark that as `last_gen_end_time` and flip some booleans so that the manager can give it out again (eventually recorded in `last_sim_given_time`)

### DIFF
--- a/libensemble/tools/fields_keys.py
+++ b/libensemble/tools/fields_keys.py
@@ -1,21 +1,22 @@
 """
-Below are the fields used within libEnsemble
+Below are the fields used within libEnsemble (All times are since epoch)
 """
 
-libE_fields = [('sim_id', int),             # Unique id of entry in H that was generated
-               ('gen_worker', int),         # Worker that (first) generated the entry
-               ('gen_time', float),         # Time (since epoch) entry (first) was entered into H from a gen
-               ('last_gen_time', float),    # Time (since epoch) entry was last requested by a gen
-               ('given', bool),             # True if entry has been given for sim eval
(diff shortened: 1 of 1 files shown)


wildsm:
This looks great.

docs/data_structures/history_array.rst
```diff
@@ -85,32 +85,22 @@ The full list of these reserved fields is given below.
 * ``given`` [bool]: True if this ``gen_f`` output been given to a libEnsemble
```
jlnav, line 85: This file still needs to be adjusted, right?

jmlarson1: I have updated this file.

jlnav:
Good feedback from Angel, I like the current state of ``fields_keys.py``. Understanding the "story" of a point is (I think) more intuitive now

jlnav (approved):

shuds13 (approved):

PR #761, opened 2022-04-01
Try fixing tasmanian install via patching setup.py

jlnav:
Should probably open a PR upstream to fix this issue. Setuptools got updated to experimentally look for packages by default, and fail if a package doesn't adhere to its standards...


jlnav:
This was an attempt, but not the best solution by far

PR #766, opened 2022-04-07
Bugfix/restore ax coverage

jmlarson1:


jmlarson1:
@shuds13 Pulling this in because it solved the ax error. But now there is an error with macOS build of gpytorch. Let's wait a few days and see if that is resolved upstream.

PR #768, opened 2022-04-07
Feature/forces gpu

shuds13:


jmlarson1 (approved):

libensemble/tests/scaling_tests/forces/forces_gpu/run_libe_forces.py
```diff
@@ -0,0 +1 @@
+../forces_simple/run_libe_forces_simple.py
```
jmlarson1, line 1: Why is the `gpu` file a symlink to the `simple` file?

jlnav (approved):

Issue #759, opened 2022-03-21
Forces structure (code and docs)


shuds13:
My proposal. This is mostly re-arranging what we already have. Use a new PR.

1 - Extracting common code of forces sub-dirs. Prob forces.c itself -e.g.

   scaling_tests/forces
     - forces_app/  (contains forces.c/build.sh).
     - forces_simple/    
     - forces_balsam/
     - forces_funcx/
     - forces_adv/  (the original forces)  
   
where forces_(simple/balsam/funcx) would use the simplified forces from the tutorial (the examples/tutorials directory would also link to that).

Maybe instead of forces_app/ have common/  if there is more can put in there like sim_f - or have other dirs override
     
2 - How do we reference in the docs - could we put whats in the balsam_forces/readme in a tutorial? Potentially have 3. In docs:
 
 tutorials/forces
    - 1. Simple forces 
    - 2. Remote via Balsam (poss funcX) 
    - 3. Advanced - polling/killing/outputfile - e.g. the old one.
  
  Also - balsam_forces - perhaps do a video (as it involves local/remote machine work and globus setup).

jlnav:
Closed by #775

PR #775, opened 2022-04-19
Refactor/reorganize forces

jlnav:
Addresses #759 .

Pulls into feature/balsam7 since this also refactors the new test in that PR.

TODO: Test new version of balsam forces with Theta.


shuds13:
Path to forces in define_apps.py will need to be ...forces_app/forces.x

Maybe balsam_forces -> forces_balsam (and same for funcx) to be consistent with others naming.

shuds13:
Why are imports inside the function in sim_f?

shuds13:
I think the executors/__init__.py still needs attention. If for example you do not have "balsam-flow" installed then it will error out on 
`pkg_resources.get_distribution("balsam-flow")`
at least it does for me. The exception catches but "balsam" (as in balsam2) does not get imported.  I think the two imports need separate catches. 
Then I also wonder if conflicts are possible if both are present and imported. As they are notl separate namespaces. Both import 'balsam'. What do we want in that scenario.

jlnav:
> Path to forces in define_apps.py will need to be ...forces_app/forces.x
> 
> Maybe balsam_forces -> forces_balsam (and same for funcx) to be consistent with others naming.

define_apps.py fixed.

So it isn't "forces" "forces" "forces" repetitively, underneath the main forces folder, thoughts on the folders just being, e.g. balsam, advanced, app, simple, funcx ?

jlnav:
> Why are imports inside the function in sim_f?

This is an artifact of copy-pasting the funcx variant. It looks odd but I wouldn't mind keeping it as-is since in the future I'll work on an update to submit that sim_f remotely also via Balsam

shuds13:
File "/home/shudson/libensemble/balsam2_process/libe1/libensemble/tests/scaling_tests/forces/balsam_forces/forces_simf.py", line 29, in run_forces_balsam
    workdir = "sim" + ["libE_info"]["H_rows"] + "_worker" + str(libE_info["workerID"])

PR #778, opened 2022-04-25
Adds libE_specs option for multiprocessing start method

jlnav:
Set ``libE_specs['mp_start_method']`` to  ``'spawn'``, ``'fork'``,  or ``'forkserver'`` in a calling script.

If this option isn't provided, we still override the start option on macOS to ``'fork'``. Otherwise we leave the default.


shuds13 (approved):

jmlarson1:
If possible, let's try to cover `set_mp_method`

PR #782, opened 2022-04-26
Black/string normalization

jmlarson1:


jmlarson1:
@jlnav The most useful reason to do this (in my mind) is making it easier to `grep` through files. Given the extensive list in `force-exclude` section of `.black`, this PR doesn't help much.

What do you think about closing this until we reduce the `force-exclude` list further?

jlnav:
>  What do you think about closing this until we reduce the force-exclude list further?

I'm fine with that. It can be one of the last changes we make

PR #787, opened 2022-04-27
Black examples

jmlarson1:
@jlnav Do you like the black formatting of the jupyter notebooks as well?


jlnav (approved):
I'm actually very impressed it could format jupyter notebooks!

PR #794, opened 2022-04-27
Add non-blocking receive option

shuds13:


libensemble/gen_funcs/persistent_uniform_sampling.py
```diff
@@ -115,3 +115,41 @@ def persistent_request_shutdown(H, persis_info, gen_specs, libE_info):
...
+        received = False
+        spin_count = 0
+        while (not received):
+            tag, Work, calc_in = ps.recv(blocking=False)
+            if tag is not None:
+                received = True
+            else:
+                spin_count += 1
+
+        assert spin_count > 0, 'Expected spin_count > 0 in non-blocking loop'
```
jmlarson1, line 150: @shuds13 Why would this have to be zero? What if the first (non-blocking) receive returns Work?

jlnav (approved):

libensemble/gen_funcs/persistent_uniform_sampling.py
jmlarson1: I guess that we've seen some weird timing stuff (mostly on GitHub). It would seem possible that `ps.recv` could be "slow" sometime and the sim_f could be fast, no?

PR #798, opened 2022-04-28
replace 0.8.0+dev mentions with new version

jlnav:


shuds13:
We did not do it yet - make sure we get the one starting "Users interested in updating "

PR #816, opened 2022-05-06
Feature / any calc status

jlnav:
Addresses #808 

Simple enough series of adjustments so users can return whatever they'd like for a calc status to log into libE_stats.

_However!_ This isn't supported for persistent functions because of (at a minimum, and probably elsewhere too)
logic in ``_update_state_on_worker_msg()``. So custom calc statuses will likely only be used within common sim_fs.


jmlarson1:
In this branch, can we group the entries in `libensemble/message_numbers.py` to those that are calc_statuses, as opposed to other types of message numbers?

jlnav:
Should this be going into the release branch instead?

jmlarson1:
> Should this be going into the release branch instead?

Everything has been going into `develop`. After it's pulled in, I just pull into the release then delete the branch (without waiting for all tests to pass).

I'm open to any procedure.

jmlarson1 (approved):

shuds13:
Has docs/data_structures/calc_status.rst been updated?

shuds13 (approved):

PR #830, opened 2022-05-17
Add `if __name__ == "__main__":` blocks to non-extra regression tests

jlnav:
Lets see what happens. Should also help with enabling windows testing?

- [ ] Remove `osx_set_start_method()` ?
- [ ] Add pytest-parallel as testing dependency?


jlnav:
Going to redo this a bit more _gently_ after our discussions

PR #834, opened 2022-05-18
Feature / parse PBS nodefile in env_resources

jlnav:
While other schedulers set some environment variable to contain the actual nodelist for a job, PBS has a ``PBS_NODEFILE`` environment variable, "The filename containing a list of vnodes assigned to the job."

Adds an additional ``env_resources.py`` nodelist entry for PBS that reads and processes the list of vnodes within the file pointed to by ``PBS_NODEFILE``.


shuds13 (approved):
We should add a submission script, but this is working for me in reading nodelists.

Issue #839, opened 2022-05-23
Separate regression tests into problem examples and artificial tests


jlnav:

jmlarson1:
Resolved by #855

PR #840, opened 2022-05-23
Refactor/separate reg tests

jlnav:
Addresses #839


jlnav:
Probably going to come back to this later

PR #841, opened 2022-05-24
Renaming/persistent sampling

jmlarson1:


shuds13:
Might want to consider making the function names more consistent. I dropped the persistent_ prefix from the functions I added as its already in the module name.

jmlarson1:
> Might want to consider making the function names more consistent. I dropped the persistent_ prefix from the functions I added as its already in the module name.

I'm not sure. This could be a larger renaming if desired (in a future PR).

PR #852, opened 2022-06-01
Possible cleanup

jmlarson1:
Feel free to accept/reject this cleanup commit.

### DIFF
--- a/libensemble/comms/comms.py
+++ b/libensemble/comms/comms.py
@@ -25,8 +25,6 @@ access and monitoring (for persistent gens):
 from abc import ABC, abstractmethod
 from time import time
 from threading import Thread
-
-# from multiprocessing import Process, Queue, Value, Lock
 from multiprocessing import Process, Queue
 from traceback import format_exc
 import queue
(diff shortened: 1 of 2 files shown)


shuds13:
I was keeping dynamic count lines around as was contemplating using them if nworkers is not passed. We just dont have a need/use-case at the moment, but I could remove them to keep it clean - I've got the dynamic variant here https://github.com/shuds13/libensemble/blob/experimental/keep_dynamic_worker_count/libensemble/comms/comms.py

As for the name==main line, I'm a bit ambivalent. I'm wondering if putting it there would make someone think it needs to be there.

jmlarson1:
> As for the name==main line, I'm a bit ambivalent. I'm wondering if putting it there would make someone think it needs to be there.

I understand the commented lines. 
But the name==main was not added by me. I just moved it down from its original location. Can it be removed altogether?

jlnav:
> I understand the commented lines. But the name==main was not added by me. I just moved it down from its original location. Can it be removed altogether?

Probably, but keep in mind I have a branch open that's adding them to most tests

jmlarson1:
These edits undo what Steve was intentionally doing. Closing.

Issue #850, opened 2022-05-31
Organize libE_specs


jmlarson1:
There are many libE_specs and the wall of text should be organized for easier reading:

https://libensemble.readthedocs.io/en/main/data_structures/libE_specs.html

jmlarson1:
Resolved by #853

PR #853, opened 2022-06-03
A first re-org pass

jmlarson1:
In response to issue #850, thoughts on these high-level categories here? Organize alphabetical in the lower categories?

### DIFF
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -36,7 +36,6 @@ libEnsemble is primarily customized by setting options within a ``libE_specs`` d
             If False, the manager avoid this moderate overhead.
             Default: True
 
-
         Directory management options:
         -----------------------------
         'ensemble_dir_path' [str]:
@@ -79,7 +78,6 @@ libEnsemble is primarily customized by setting options within a ``libE_specs`` d
             Copy this directory and its contents for each generator-instance specific directory.
             If not using calc directories, contents are copied to the ensemble directory.
 
-
         Profiling/timing options:
         -------------------------
         'profile' [bool]:
@@ -91,7 +89,6 @@ libEnsemble is primarily customized by setting options within a ``libE_specs`` d
             Prevents user functions from overwriting protected libE fields, but requires
             moderate overhead.
             Default: True
-
 
         TCP only options:
         -----------------
@@ -110,7 +107,6 @@ libEnsemble is primarily customized by setting options within a ``libE_specs`` d
             a local Python path, calling script, and manager/server format-fields for manager_ip,
             manager_port, authkey, and workerID. nworkers is specified normally.
 
-
         Options for history management with persistent workers:
         -------------------------------------------------------
         'use_persis_return_gen' [bool]:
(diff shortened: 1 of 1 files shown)


jlnav:
Since nothing here is literal-included, could the sub-categories be divided into rst sub-headings so it's easier to see them on a glance? Thoughts?

jmlarson1:
I'm not sure this is possible. When I tried `**header**` or 

````
header
------
````
they rendered with the "`-`" and "`*`" (not formatted).  Can you show me what you are thinking?

**Comment by shuds13** [comment]
I think the categories are fine. Maybe add a 'general' sub-heading for those that dont fit in other categories. And maybe add a line before the schedular/resource_info section to say these are for sub-dictionaries to above options.

**Comment by jmlarson1** [comment]
@shuds13 I've implemented your suggestions. I'm not sure why the first block of text (starting line 8) is formatted differently than the second block of text (starting line 170). Not sure it matters right now.

File: docs/data_structures/libE_specs.rst [hunk 1]
```diff
@@ -128,12 +146,14 @@ Specifications for libEnsemble::
         'resource_info' [dict]:
             Provide resource information that will override automatically detected resources.
             The allowable fields are given below in 'Overriding Auto-detection'
-            Note that if ``disable_resource_manager`` is set then
+            Note that if "disable_resource_manager" is set then
             this option is ignored.
         'scheduler_opts' [dict]:
             A dictionary of options for the resource scheduler.
             See 'Scheduler Options' for more.
```
wildsm, line 153: See 'Scheduler Options' for more options.

(add " options")

docs/data_structures/libE_specs.rst
```diff
@@ -144,32 +164,32 @@ See options for :ref:`built-in scheduler<resources-scheduler>`.
 Overriding Auto-detection
```
wildsm, line 164: Overriding Auto-Detection

docs/data_structures/libE_specs.rst
```diff
@@ -24,6 +28,19 @@ Specifications for libEnsemble::
             Whether libEnsemble should save the states of H and persis_info on
             aborting after an error.
             Default: True
+        'worker_timeout' [int]:
+            When libEnsemble concludes and attempts to close down worker processes,
+            the number of seconds after which workers are considered timed out. Worker
+            processes are then terminated. multiprocessing default: 1
```
jmlarson1, line 34: Is this `1 second` or something else?

jlnav:
> Can we add "only" to the `nworkers` statement about tcp/local modes?
> 
> Also, can we organize within each grouping?

What subcategories are you considering?

jmlarson1:
> > Can we add "only" to the `nworkers` statement about tcp/local modes?
> > Also, can we organize within each grouping?
> 
> What subcategories are you considering?

I was thinking (for example) to have `Directory management options:` be alphabetical in order (or maybe `use_worker_dir` to be first.

shuds13:
It seems the line

The following describe the dictionary options within ``libE_specs``.

was added but then removed?

jmlarson1:
I think it was inadvertently dropped in ceba837e 

I added it back

PR #855, opened 2022-06-06
Another approach to separating some regression tests

jlnav:
Once more, with feeling!

Probably need to check out the branch locally to effectively eyeball how the two directories look

Addresses #839


shuds13:
i think this captures pretty well functionality versus mini examples. My only doubt is when there is a functionality test that may make a good example for the user and is not covered in regression tests. The main one I thought of was `test_persistent_uniform_gen_decides_stop.py` as I think this is the only test where gen triggers the shutdown.

shuds13 (approved):

Issue #797, opened 2022-04-28
Try to support 'spawn' processes


shuds13:
At very least executor and resources are set on the parent process (manager) and get copied by the fork to worker processes. 

We can either change this by setting up executor/forces later, or explicitly copy the data structures.

This is necessary for Windows support and some pytorch use-cases

PR #857, opened 2022-06-16
Refactor/dreaded line

jlnav:
Lets see how this plays out

Addresses #797


libensemble/libE.py
```diff
@@ -100,6 +100,46 @@
             y:
                 type: float
 
+On macOS and Windows, when using multiprocessing (``local`` comms), you may need to place most
```
shuds13, line 103: I would not say this is about leaked semaphore warnings or pytorch. Rather, without it, is isn't going to work. 

I would say more like the warning line for tests - Something along the lines of - If using local comms and the process start method is 'Spawn' (used by Windows, and the default on macOS since python 3.8), then you must use `if __name__ == "__main__:". You could add the that bracketed part - also the startmethod required by pytorch, perhaps.

shuds13 (commented):
I left the one comment - other than that I'm happy with this.

libensemble/libE.py
shuds13: If you want, could also point out the alternative of setting start method to 'fork' on macOS.

jmlarson1 (approved):

docs/running_libE.rst
```diff
@@ -73,6 +73,10 @@ system (e.g. Theta/Summit), allowing the whole node allocation for
 worker-launched application runs. In this scenario, make sure there are
 no imports of ``mpi4py`` in your Python scripts.
 
+Note that on macOS and Windows, the default multiprocessing method is ``"spawn"`` instead
```
shuds13, line 76: Maybe specify macOS default since version 3.8.

shuds13 (approved):
Approve on passing tests

Issue #574, opened 2021-01-05
Experiment with Windows on GitHub Actions


jlnav:

jmlarson1:
@jlnav Should we close this for now?

jlnav:
> @jlnav Should we close this for now?

I believe so. I'm relatively unfamiliar with how we'd install most of the non-Python packages on windows, and debugging this process might take lots of time for (I think) little benefit.

shuds13:
@jlnav @jmlarson1 
But we could test Windows without the non-python packages. Make sure core libE works.

jmlarson1:
Or, in other words, that @jlnav instituted extra tests, can we run the nonextra tests in Windows?

jlnav:
7/27/22 - Log for investigating installing MPI on Windows:

- msmpi
   -  Installable from conda, but compilers not included
   -  no longer maintained, according to discussion on msmpi feedstock
   -  https://docs.microsoft.com/en-us/archive/blogs/windowshpc/how-to-compile-and-run-a-simple-ms-mpi-program
   - Trying path adjustment, compilation, and linking from above site

- mpi4py
   - no insight on compiling since mpi4py primarily concerned with MPI-Python interface

will be editing as I find more

jlnav:
Closed by #869 !!!

PR #869, opened 2022-07-12
Testing/windows

jlnav:
Addresses #574


install/testing_requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov==2.12.1
 pytest-timeout==1.4.2
 mock==4.0.3
```
jmlarson1, line 6: Are these all supposed to be hard-coded to specific versions?

libensemble/tests/unit_tests/test_executor.py
```diff
@@ -781,6 +812,7 @@ def test_non_existent_app():
         assert 0
 
 
+@pytest.mark.extra
```
jmlarson1, line 815: Is putting this decorator before every function the "right way" to do this?

jmlarson1 (approved):

libensemble/tests/unit_tests/test_executor.py
jlnav: We made it ourselves, I put it before tests that fail on windows so they don't get run. Most of the executor tests would be very difficult, if not impossible, to get running on windows since `mpicc` isnt available.

The decorator is produced by pytest based on the settings in `unit_tests/conftest.py`

shuds13 (approved):

PR #872, opened 2022-07-15
Re-enable PETSc tests

shuds13:


shuds13:
Installing dragonfly from git as do not know when the fixed code will be released.

jlnav (approved):

PR #876, opened 2022-07-20
[Experimental] user func field decorators

jlnav:
Experimental approach to label and assign fields to user functions instead of listing them in every set of sim_specs or gen_specs that uses that function.

I've also verified that once the function has been imported in a calling script, the fields can still be adjusted (for those cases where user functions check if the user has defined some optional field)


jlnav:
Lets come back to this later

PR #877, opened 2022-07-22
Update manager_poll interface

shuds13:
- Replace manager poll string output with types - decision and look at
- Should we support both for deprecation
- Add/test function manager_kill_received(), returns boolean.


jlnav:
Maybe I'm just not thinking clearly, but we'd hypothetically somehow need to support `manager_poll()` returning both an integer tag and a string?

I guess add an argument like `return_tag=True` for backwards compatibility? But I think it looks weird.

I'd prefer the breaking change, personally. Thoughts?

shuds13:
> we'd hypothetically somehow need to support `manager_poll()` returning both an integer tag and a string?

Yes, that would be a problem. If we go with this change, should probably not deprecate.

jlnav (approved):

jmlarson1 (approved):

jmlarson1 (approved):

PR #880, opened 2022-07-26
Feature/stay active and cancel func

shuds13:
Add stay_active option to PersistentSupport

Add cancel_points convenience function to PersistentSupport

Add persistent sampling cancellation test.

**API terminology**

stay_active (a one-way flip) could altrnatively be called (some with T/F flip): 

- keep_status / keep_state / preserve_state
- new_data (where new_data=True counts as new points (gen) / new results (sim))
- set_inactive or ready2recv / set_ready2recv
- Instead a 3-way variable None/True/False where None either does not change or does the manager default - while others set either way.
-  [x] keep_state

cancel_points could be:
- cancel_sim_ids, req_cancel_points
- [x] request_cancel_sim_ids


shuds13:
Possible uses of stay_active

- Cancelling points - is in the new function
- Other changes to existing history values
- Signals to manager (e.g. I want more resources - or I want to release resources)
- Worker sends points to manager but  is still working and will send more.


Note that the option active_recv currently keeps a worker in active mode always - that could in theory  be changed with this option - but should it be?

jmlarson1 (approved):
Either in this PR, or in a soon-to-follow PR, we need to document the now 3 ways of cancelling sim_ids.
- The surmise approach
- Sending back new points along with marking `cancel` for existing points that the worker is aware of
- This new approach (first send back cancel requests for some sim_ids, and let the manager know that more is to come)

shuds13:
I will update docs. 

The surmise use of an active-receive worker allowed cancel points without this change, but I think of that mode more for irregular communication patterns for giving/receiving points. Other codes may be happy with a ping-pong pattern, but want to cancel points. I will make  the surmise example use this function anyway.

shuds13:
Ive removed keep_state argument from the send_recv function, as I don't see a likely use-case.

docs/function_guides/generator.rst
```diff
@@ -165,6 +161,28 @@ by a user function, otherwise it will be ignored.
...
 
+Modification of existing points
+-------------------------------
+
+To change existing fields of the history array, the generator can initialize an output
+array where the *dtype* contains the ``sim_id`` and the fields to be modified (in
+place of ``gen_specs["out"]``), and then send to the manager as with regular
+communications. Any such message received by the manager will modify the given fields
+for the given *sim_ids*. If the changes are separate from the generation of new points
+and should not mark the generator as ready to receive new points, then send with the
```
jlnav, line 172: maybe instead of "If the changes are separate from the generation of new points
and should not mark the generator as ready to receive new points,"

"Since these changes don't correspond with newly generated points, then the generator needs to communicate to the manager that it isn't necessarily ready to receive completed evaluations. Send to the manager with ``keep_state`` set to True."

or something like this

jlnav (approved):
Looks good!

docs/function_guides/generator.rst
jmlarson1: @jlnav Shouldn't the first "Since" be an "If"?

jmlarson1 (approved):

docs/function_guides/generator.rst
shuds13: Updated

Issue #883, opened 2022-08-03
Mock Balsam and funcX modules for proper testing


jlnav:

jlnav:
Closed by #886 and #890

PR #890, opened 2022-08-18
Testing/mock balsam2

jlnav:
Unit tests for our new Balsam executor. I branched off the funcx branch so I could easily reference how I used mock, so the funcx branch should probably be pulled in first.

Addresses #883


jlnav:
Tests passing, just need to add additional tests to increase coverage

shuds13 (approved):

PR #891, opened 2022-08-22
[WIP] Refactor/aposmm opt setup

jlnav:


jlnav:
This currently doesn't appear to work on windows, with **rsopt**, due to importing aposmm_localopt_support causes the module to try to access optimizers from init.py before they've been written anywhere.

jmlarson1 (approved):

jlnav:
I actually think these changes are still good for **libEnsemble**. But I'll keep investigating alternatives for rsopt.

jlnav:
Oh yeah, this is probably still ready to review and pull in. RC now reads from the file on initialization if it exists, and only overwrites the file if the setter's contents differ from the file contents (to avoid constantly overwriting the file)

shuds13:
I think we can simplify this, by not ever writing the file ourselves. As this works when running libEnsemble, including on Windows (at least for me), I think a better approach is for _rsopt_ (or anyone for that matter) can produce the file. Then in `__init.py__`, we can check and read the file if it exists, and otherwise just use the original method.  But we should agree this with @cchall.

jlnav:
>  Then in __init.py__, we can check and read the file if it exists, and otherwise just use the original method. But we should agree this with @cchall.

Thanks for the reminder. I'll look into those changes then ping Chris (unless being pinged here is satisfactory?)

cchall:
Thanks for implementing this!
I tried it out and this solution will definitely work fine for rsopt.

I do have a minor quibble with the fact that `RC` is both defined and called in `gen_funcs/__init__.py`. This means that I can't use `import libensemble.gen_funcs` in a straightforward manner to find the path to write .opt_modules.csv to. 

In general it might also be nice if the filename that `RC._csv_path` expects were defined somewhere else so that users can import it from libensemble rather than having to hardcode it. Though this is a very minor thing.

jlnav:
> Thanks for implementing this! I tried it out and this solution will definitely work fine for rsopt.
> 
> I do have a minor quibble with the fact that `RC` is both defined and called in `gen_funcs/__init__.py`. This means that I can't use `import libensemble.gen_funcs` in a straightforward manner to find the path to write .opt_modules.csv to.
> 
> In general it might also be nice if the filename that `RC._csv_path` expects were defined somewhere else so that users can import it from libensemble rather than having to hardcode it. Though this is a very minor thing.

Hi Chris, thanks for the feedback! I may be misunderstanding, but in the current iteration the path shouldn't need to be hardcoded; the following works for me:

```
import libensemble.gen_funcs
with open(libensemble.gen_funcs.rc._csv_path, "w") as f:
   f.write("nlopt")
```

I can simplify this of course by adding `csv_file = rc._csv_path` to the bottom of `__init__.py` so that the path is simpler: `libensemble.gen_funcs.csv_file`. This path should be valid on any system where libensemble is installed. Thoughts?

cchall:
I ended up modifying where I do the aposmm opt setup to just be run as part of rsopt's general startup procedure so I can just call `import libensemble.gen_funcs` and get the path. So I don't think it's necessary to make any changes here.

The issue I was previously running into is that I was trying to keep the aposmm opt setup in the same location, with rsopt's local optimizer setup so that it is only invoked when needed. The problem then is that `import libensemble.gen_funcs` cannot be called without causing an error. Though some hardcoding can get around this:
```
import libensemble
_libensemble_path = pathlib.Path(libensemble.__file__)
with open(_libensemble_path.parents[0].joinpath('gen_funcs/.opt_modules.csv'), 'w') as ff:
     ff.write(','.join(available_opt))
```

jlnav:
Okay, fair enough (I think), I'll leave the path determination alone.

Maybe I'm still misunderstanding; what type of error are you getting when `import libensemble.gen_funcs` occurs? I'm puzzled how that happens if every Python process can access the module, unless there's an issue with the auto-setting of opt modules from the file path when the module is imported.

jlnav:
Alternatively, a reproduce would be appreciated if possible!

shuds13 (approved):

PR #902, opened 2022-09-14
Refactor/fstrings

jlnav:
Changes most `.format()` function calls to instead use f-strings.

Besides f-strings being newer and more well supported, they're actually faster:  https://www.scivision.dev/python-f-string-speed/


wildsm:
As discussed today, more context for flynt: https://github.com/ikamensh/flynt
Probably could consider as part of an action but might need to consider whether tests cover formatted string things (e.g., as noted in Dangers of conversion in above URL).
A great tool to in any case to help our flow.

shuds13 (approved):

jlnav:
I'm thinking I'll mention flynt as part of a release process tool in the future

Issue #882, opened 2022-07-29
Make num_resource_sets the recommended way to use zero resource gen


shuds13:
A specified zero resource worker is only necessary when a particular worker must be used (e.g. gen must run on worker 1). For most users this is not the case, so restructure docs to recommend the `num_resource_sets` _libE_specs_ option instead (and/or command line option `n_resource_sets`) instead of the `zero_resource_worker` option.

This can be used where sensible in examples, docs, tutorials, presentations. Some tests will still use zero_resource workers.

The `nsim_workers` command line option (_parse_args_) could also be changed to set this option - discuss?

`zero_resource_workers` will continue to be supported (where allocation functions support). This can also be extended.

jlnav:
Closed by #905

PR #905, opened 2022-09-20
Use num_resource_sets instead of zrw in most cases

shuds13:
See issue #882

- [x] Change selected tests
- [x] Change in docs (examples tutorials)
- [x] Change `nsim_workers` command line option (_parse_args_)  to set num_resource_sets

Also: Add or update tests in templater


jlnav (approved):
Looks good!

PR #906, opened 2022-09-20
Adding another pointer to contributing doc

jmlarson1:


wildsm (changes requested):
What about
We welcome others to Contribute_ to libEnsemble
->
Contributions_ to libEnsemble are welcome
(and change targets

PR #917, opened 2022-11-09
Deprecate Py 3.7, Add Py 3.11, various adjustments

jlnav:


.github/workflows/ci.yml
```diff
@@ -144,7 +139,7 @@ jobs:
             pip install DFO-LS
 
             pip install mpmath
-            pip install ax-platform
+            pip install ax-platform==0.2.8
```
shuds13, line 142: We should note where we have done this on the next release issue, so we can see then if works with latest

PR #921, opened 2022-11-14
Retire LegacyBalsamMPIExecutor

jlnav:
Looks like the Balsam team removed all their Balsam 1 releases and docs:

https://balsam.readthedocs.io/en/master

https://github.com/argonne-lcf/balsam/archive/refs/tags

This removes the LegacyBalsamExecutor and its tests. On the off-chance that a user _really_ wants to use Balsam 1 and is able to install it, I'm thinking they can use libE up to v0.9.3.

Jobs are failing due to missing changes in #917


libensemble/executors/__init__.py
```diff
@@ -1,4 +1,10 @@
+import importlib
 from libensemble.executors.executor import Executor
 from libensemble.executors.mpi_executor import MPIExecutor
 
-__all__ = ["Executor", "MPIExecutor"]
```
shuds13, line 4: Is this just old balsam?

libensemble/tests/deprecated_tests/standalone_executor_tests/simdir/my_simtask.c
```diff
@@ -1,42 +0,0 @@
-#include <unistd.h>
```
shuds13, line 1: Is tihs just usd for Balsam?

libensemble/tests/deprecated_tests/test_old_aposmm_sim_dirs.py
```diff
@@ -1,112 +0,0 @@
-# """
```
shuds13, line 1: balsam?

shuds13 (commented):
I want to check we are only removing balsam stuff - or we've all agreed for other stuff. Also, we want to make sure the new balsam still works before we pull this in.

PR #925, opened 2022-11-18
Feature/pyproject toml

jlnav:
Reorganizes setup.py and many utility configuration scripts into a new `pyproject.toml`, as is typical in modern python projects. Also some cleanup. _Probably_ still a WIP


shuds13:
Have you checked spack install still works with this?

jlnav:
> Have you checked spack install still works with this?

Good point, I need to do that

setup.py
```diff
@@ -1,127 +1,3 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-# This is for setting up libEnsemble, license and details can be
-# found at https://github.com/Libensemble/libensemble/
-
-"""libEnsemble
```
shuds13, line 6: Put docstring back in - used in PyPI

jlnav:
Some links I looked at originally regarding pyproject.toml:

https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/
https://peps.python.org/pep-0518/
https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for
https://betterprogramming.pub/a-pyproject-toml-developers-cheat-sheet-5782801fb3ed

Another benefit I didn't mention is if other developers want to build/contribute to libE _without using setuptools or pip_ (which is a thing you can do nowadays!), they can do so.

We _may_ also be able to move pytest and coverage configs into the pyproject.toml too, but I won't do that in the short term

PR #928, opened 2022-12-02
Add GPU detection module

shuds13:
Add GPU detection module and use in dynamic resources examples.

- Needs to work on Summit
- Test across systems
- Add unit tests with resource_info options - need a test with `resource_info['gpus_per_node']`
- Update forces_gpu example / tutorial
- document changes - including dynamic resources overview / worker_resources ivars (gpus_per_...)
- Before merging - check prints are removed.

Discuss: 

For querying GPU info - we can get `gpus_per_node` (absolute value) `gpus_per_rset`, but what about gpus available to this worker (in total / on a node) - which depends on `rsets` and `gpus_per_rset`. How to use consistent terminology. This should be reviewed for existing variable names also (e.g. Clarify total v assigned to this worker - and total v per node).

e.g. slot_count  "The number of slots per node if even_slots is True, else None."
should be: "The number of slots per node assigned to this worker if even_slots is True, else None."

Names should distinguish in consistent way.

Main: https://libensemble.readthedocs.io/en/main/resource_manager/worker_resources.html
This branch: https://libensemble--928.org.readthedocs.build/en/928/resource_manager/worker_resources.html

We could also automate setting eg.,~ `CUDA_VISIBLE_DEVICES` directly to the available GPUs - similarly could set `num_procs` to the total GPUs available OR `procs_per_node` in a similar way.

### DIFF
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -188,6 +188,8 @@ The allowable ``libE_specs["resource_info"]`` fields are::
 
     "cores_on_node" [tuple (int, int)]:
         Tuple (physical cores, logical cores) on nodes.
+    "gpus_on_node" [int]:
+        Number of GPUs on each node.
     "node_file" [str]:
         Name of file containing a node-list. Default is "node_list".
     "nodelist_env_slurm" [str]:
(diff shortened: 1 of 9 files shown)


shuds13:
For the consistent terminology - these are all under worker_resources - so perhaps all values here should be only for THIS worker. `gpu_per_node` actually is just copied from global_resources, so perhaps it should be a part of that?

PR #931, opened 2022-12-09
fixing some dep installs as usual

jlnav:


libensemble/history.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, alloc_specs, sim_specs, gen_specs, exit_criteria, H0):
             # - dtypes aren't iterable, but you can index into them
             # - must split out actual numpy type if subdtype refers to sub-array
             # - then convert that type into a python type in the best way known so far...
+            # - we need to make sure the size of string types is preserved
```
jmlarson1, line 60: Can we make a simple test of this?

jlnav:
Windows builds are timing out - this will be tough to debug, so will do it separately

PR #936, opened 2022-12-13
Expanding heFFTe options

jmlarson1:
@mkstoyanov Can we use this PR for any additional heFFTe options?

### DIFF
--- a/libensemble/tests/regression_tests/test_heffte.py
+++ b/libensemble/tests/regression_tests/test_heffte.py
@@ -27,8 +27,10 @@ if __name__ == "__main__":
     fixed = ["mpirun -np 4 ./speed3d_c2c fftw double 128 128 128"]
     arg1 = ["-no-reorder", "-reorder"]
     arg2 = ["-a2a", "-a2av", "-p2p", "-p2p_pl"]
+    arg3 = ["-ingrid 4 1 1", "-ingrid 2 2 1"]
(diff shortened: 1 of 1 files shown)


mkstoyanov:
I don't think we need to run all 32 cases, that's why I had left one as a comment rather than a used variable.

There are a few more options, but then we are getting into GPU territory and cases where the size of the comm should really be more than 4. What we have now shows clearly how to add the other options, this would be done by the user for the specific use-case.

We have the framework and we can mark progress on the issue. The next step is to find a better way to search through the cases without running them all.

Miro

On Tue, 2022-12-13 at 09:50 -0800, Jeffrey Larson wrote:

@jmlarson1<hxxps://github.com/jmlarson1> requested your review on: #936<hxxps://github.com/Libensemble/libensemble/pull/936> Expanding heFFTe options.

mkstoyanov (approved):

jmlarson1:
@mkstoyanov  Thanks for the quick response. 

Given that the runtimes are all less than 1 second, to make the problem somewhat more difficult (if desired), do you have any specific larger problem sizes?

mkstoyanov:
@jmlarson1 since autotuning is so machine specific, it makes little sense to make a problem "harder" for the CI.

The "real" problem has to come form the user and that has to cover much larger sizes, more comms, more variations for the `ingrid` and `outgrid` parameter. Anything that we can put in a proof-of-concept only.

PR #937, opened 2022-12-13
Re-add and allow alloc_specs["out"]

jlnav:


libensemble/specs.py
```diff
@@ -173,6 +173,12 @@ class AllocSpecs(BaseModel):
     A user-data dictionary to place bounds, constants, settings, or other parameters for customizing the allocation
     function
     """
+
+    out: Optional[List] = []
+    """
+    Deprecated. Optional list of 2 or 3-tuples corresponding to numpy dtypes. Previously used to
```
jmlarson1, line 179: Any reason why deprecated? (Because we consider sim_specs['in'] and gen_specs['in'] when constructing H, along with the ['out'] items?)

jlnav: @shuds13 may have some insight (although I believe I pushed for the deprecation originally).

Basically two reasons:

1. Where ``sim_specs["out"]`` and ``gen_specs["out"]`` describe fields that are output by the sim and gen, ``alloc_specs["out"]`` never really described fields that are output by the allocation function, only fields that needed inclusion in H but didn't belong in ``sim_specs["out"]`` or ``gen_specs["out"]``

2. Now that I check again, and related to the first point: ``alloc_specs["out"]`` isn't used in _any_ allocation function!

So how do we specify extra fields that go _into_ a function but not out of it? If they're not generated (coming out of the gen), they're coming from an H0. Where we now allow users to put whichever fields they want!

jmlarson1 (approved):

PR #959, opened 2023-01-12
Debugging when work is not a dictionary

jmlarson1:


jmlarson1:
Is this the correct way to raise this error?

jmlarson1:
There are various intermittent errors. For two 

https://github.com/Libensemble/libensemble/actions/runs/3903957476/jobs/6669062308
https://github.com/Libensemble/libensemble/actions/runs/3903957476/jobs/6669061312

we will add prints to the tests themselves. For the other intermittent error:
https://github.com/Libensemble/libensemble/actions/runs/3902762108/jobs/6666218712
we have to add a print to `worker.py`.

jlnav (approved):
Looks good, probably also good to add an additional internal comment to revisit before release. shouldn't need to wait for tests to pass again on pushing that commit

jmlarson1:
> Looks good, probably also good to add an additional internal comment to revisit before release. shouldn't need to wait for tests to pass again on pushing that commit

I'm hoping that https://github.com/Libensemble/libensemble/issues/960 will suffice.

PR #964, opened 2023-01-18
initial commit for experimenting with profiling libE

jlnav:
When the output profile files are dumped by the profiler itself, they can be visualized by `snakeviz`: https://jiffyclub.github.io/snakeviz/

Within any test: `libE_specs["profile"] = True`

Then run snakeviz on the resulting files: `snakeviz manager.prof`

This opens a browser window with the usual statistics plus visualizations of the data.

TODO: 

- [ ] See if there's a way to combine results from the workers

### DIFF
--- a/libensemble/manager.py
+++ b/libensemble/manager.py
@@ -9,7 +9,6 @@ import glob
 import logging
 import os
 import platform
-import pstats
 import socket
 import sys
 import time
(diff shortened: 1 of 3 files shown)


jlnav:
Not especially investigating this in the very-short-term I believe, closing this, just for now.

PR #967, opened 2023-01-19
Trying latest openmpi version

jmlarson1:


jmlarson1:
@jlnav This resolved the openmpi build issue, but other transient regression tests failed, but I think that's because GitHub CI is having issues today.

jlnav (approved):

PR #926, opened 2022-11-28
Refactor/type hint major modules

jlnav:
Somewhat guided by monkeytype: https://github.com/instagram/MonkeyType

PR #969, opened 2023-01-25
Ci/speedup

jmlarson1:


jlnav (approved):

jmlarson1:
@jlnav Is it easy to see if we are faster on these tests or if we lost coverage?

jlnav:
> @jlnav Is it easy to see if we are faster on these tests or if we lost coverage?

There's only one set of checks currently, so you'd have to compare the runtimes to another non-draft, recent, passed PR, e.g.: https://github.com/Libensemble/libensemble/pull/926

Where the runtimes are somewhat improved.

And unfortunately, as we discussed, the coverage checking is somewhat manual until the bot starts working for us again. So e.g. I opened the coverage report, checked fast_alloc.py, and everything there still makes sense.

Issue #935, opened 2022-12-13
Allow specifying user functions with <4 parameters


jlnav:
e.g. if a user provides a sim_f that only accepts H (input array):

```
def sim_find_sine(In):
   return np.sine(In["x"])
```

- We'd need to assume that the above function's output array still adheres to ``sim_specs["out"]``. 
- The contents of ``sim_specs["out"]`` is still necessary to build H, even though sim_specs isn't being passed into the sim_f.
- But if the contents aren't _needed_ in the sim_f, I still propose:

```
@inputs(["x"])
@outputs([("f", float)])
def sim_find_sine(In):
   return np.sine(In["x"])
```

with sim_specs simply being:

```
sim_specs = {
   sim_f: sim_find_sine
}

# or

sim_specs = SimSpecs(sim_f=sim_find_sine)
```

Of course it would still be allowed to specify more parameters in the function signature for those functions that need them. But for simpler functions it would no longer be required

jlnav:
Closed by #971 !

PR #971, opened 2023-01-26
Initial changes to allow user-functions to have <4 params, return <3 values

jlnav:
- [x] What's the simplest 1d_sampling test we could have now? :)
- [x] Documentation updates

Addresses #935


jlnav:
```
@libE_fx
def sim_f(In):
    return np.linalg.norm(In)
```

I guess to try signifying that the user_f is a "simple" evaluator.

I try unpacking a matching "item" from calc_in, then packing the result(s) back into a new H after returning.

Somewhat more complicated solution than I envisioned since functions don't have access to their own specs in/out fields!

shuds13 (commented):
Can we support dictionaries? Either different decorator name or overload.

jlnav:
Based on discussion the further-simplification-decorator can come later. The ability to register functions that accept <=4 parameters is worth merging already,

libensemble/tests/regression_tests/test_1d_super_simple.py
```diff
@@ -0,0 +1,60 @@
...
+        "gen_f": gen_f,
+        "out": [("x", float, (1,))],
+        "user": {
+            "gen_batch_size": 500,
+            "lb": np.array([-3]),
+            "ub": np.array([3]),
+        },
+    }
+
+    persis_info = add_unique_random_streams({}, nworkers + 1, seed=1234)
```
shuds13, line 51: Why do we used fixed seed here?

libensemble/worker.py
```diff
@@ -253,22 +253,27 @@ def _handle_calc(self, Work: dict, calc_in: npt.NDArray) -> (npt.NDArray, dict,
...
+                    return out
+                elif len(out) == 2:  # Out, persis_info OR Out, calc_status
+                    if isinstance(out[1], int) or isinstance(out[1], str):  # got Out, calc_status
+                        calc_status = out[1]
+                        return out[0], Work["persis_info"], calc_status
+                    return *out, calc_status  # got Out, persis_info
+                else:
+                    return out, Work["persis_info"], calc_status
+            else:
+                return np.zeros(1), Work["persis_info"], calc_status
```
shuds13, line 275: What happens with this zero? Should it be None?

libensemble/tests/regression_tests/test_1d_super_simple.py
jmlarson1: From what I see, this is the only test where we fix a seed to `add_unique_random_streams`. So we are fixing the seed just to test that part of code.

libensemble/worker.py
jlnav: Based on discussion, it probably _could_ be None, then the manager is responsible for slotting in the value

jlnav:
- [ ] Check behavior of blank specs_out-fields and returning nothing from a function

PR #979, opened 2023-02-21
Feature / adds logger functions to place logfiles in separate directory

jlnav:
`logger.set_directory("output_dir")`


jlnav:
Following the meeting, there's probably more of an interest in an overarching _workflow_ directory that itself contains the ensemble directory and each log/output file. And by default we can create new workflow directories each run differentiated by hash or date

jlnav:
`libE_specs["use_workflow_dir"] = True`

Creates a `workflow_asdf123` directory. Log-files are written here, and `persis_info`/`H` are dumped here on exception. If using an ensemble dir in another directory, copy_back will copy those files back into the workflow dir.

PR #980, opened 2023-02-22
Option to set cpus/gpus to slots in executor

shuds13:
Extension to gpu_detect branch. Executor options to set CPUs and GPUs to slots using either the method supplied in configuration or determined by the detected environment. This simplifies sim functions and makes them more portable.

There are different ways this could be specified in the arguments to submit. Current approach uses`auto_assign_gpus` and optionally `match_procs_to_gpus`, where `auto_assign_gpus` will use all GPUs assigned to current worker. Method for setting GPUs determined by either MPI runner, specified platform (via env variable LIBE_PLATFORM), or default.

1. Initial single node testing (using CUDA var resources reg. test and forces_gpu):

- [x] Laptop (inc dry_run mock mpi_runners)
- [x] Spock (works with detection - uses srun options). Can also work with ROCR_VISIBLE_DEVICES.
- [x] Crusher (works with detection - issue across GCDs not caused by this PR - known issue)
- [x] Summit (detection works but need to resolve per-task v per-node*  - num_gpus from detected platform)
- [x] Perlmutter (works with srun given - or system env var - also now works with default mpirun/CVD!)
- [x] Polaris
- [ ] Swing 
- [ ] Sunspot - currently need to set GPU/worker (and MPI rank) affinity via a bash script. Sep. branch for this.

2. Advanced testing: multiple nodes / mpi4py comms / awkward partitions.

*per-task v per-node:
Issue with using `gpu_setting_type/gpu_setting_name` pair is the meaning of the setting (e.g. settings can include gpus per node / gpus per task, not to mention tasks per gpu (e.g. on SLURM which may assign CPU tasks to have better affinity with GPUs). On Summit, the meaning of `-g` is really gpus per "resource set" (by Summit terminology) - currently interpreted in libE as gpus per task. Unfortunately, this looks like it will require more variants of `gpu_setting_type`.

Current options for gpu_setting_type:
GPU_SET_DEF = 1  # Default to MPI runner default
GPU_SET_ENV = 2  # An environment variable
GPU_SET_CLI = 3   # An MPI runner command line  option.

- [x] Add GPU_SET_CLI_GPT (gpus per task)


libensemble/executors/mpi_executor.py
```diff
@@ -142,6 +142,60 @@ def _launch_with_retries(self, task, runline, subgroup_launch, wait_on_start):
...
+        """Assign GPU resources to slots
+
+        First tries getting method from user settings, otherwise use detection or default.
+        """
+
+        resources = self.resources.worker_resources
+
+        #TODO  Add here - check for user supplied GPU option first
+
+        if self.mpi_runner.run_command == "srun":
```
shuds13, line 173: This might fit better in mpi_runner module as a member function.

shuds13: Similar jsrun setting.

shuds13:
Following discussion with @jlnav consider avoiding `slot` terminology and using `auto_assign_gpus` and maybe `match_procs_to_gpus`.

PR #982, opened 2023-02-23
Set "use_workflow_dir" to create a directory for logfiles and default ensemble dirs

jlnav:
Set `libE_specs["use_workflow_dir"] = True` for libE to create a directory to store logfiles and copied-back ensemble directories


shuds13 (changes requested):
- Make sure MPI comms produces only one workflow dir.
- Add save output function to the object interface (so works as desired).
-

shuds13 (approved):

PR #984, opened 2023-03-13
Refactoring platforms.py into pydantic classes

jlnav:
Note the validators on the base `Platform` class that apply to each of the subclasses. Note how attributes are `.`accessed.

Creating/providing a custom subclass probably needs to be tested

Also, for those functions that access attributes of the classes, maybe a class-instance could be returned to the Resources module instead for direct-use.

For _unknown_ systems, a Platform instance could be constructed on-the-spot with detected values to keep the interface the same.


libensemble/resources/platforms.py
```diff
@@ -10,91 +14,116 @@
...
-#TODO MAKE ALPHABETICAL
+class Platform(BaseModel):
+    mpi_runner: str = "srun"
+    runner_name: Optional[str]
+    cores_per_node: int
+    logical_cores_per_node: int
+    gpus_per_node: int
+    gpu_setting_type: int
+    gpu_setting_name: str
+    scheduler_match_slots: bool = False
```
jlnav, line 26: should default to True

libensemble/resources/platforms.py
```diff
@@ -10,91 +14,116 @@
...
-    "logical_cores_per_node" : 208,  # finds - check
-    "gpus_per_node" : 6,
-    "gpu_setting_type": GPU_SET_DEF,
-    "scheduler_match_slots": True,
-    }
-
 
-#TODO MAKE ALPHABETICAL
+class Platform(BaseModel):
+    mpi_runner: str = "srun"
```
jlnav, line 19: should default to mpich

jlnav: nevermind defaults not needed

libensemble/resources/platforms.py
```diff
@@ -10,91 +14,116 @@
...
-    "cores_per_node" : 104,  # finds - check
-    "logical_cores_per_node" : 208,  # finds - check
-    "gpus_per_node" : 6,
-    "gpu_setting_type": GPU_SET_DEF,
-    "scheduler_match_slots": True,
-    }
-
 
-#TODO MAKE ALPHABETICAL
+class Platform(BaseModel):
```
shuds13, line 18: We don't want default values.

shuds13:
@jlnav Are most of these commits from a merge?

jlnav:
I believe this is no longer needed to due a manual merge

PR #986, opened 2023-03-15
Rearrange specs.py libE_specs fields to match libE_specs.rst

jlnav:
And rearrange components of tcp_mgr. py for macos support


jmlarson1:
I'm guessing that it wasn't possible to group the fields like we had before:

https://libensemble--986.org.readthedocs.build/en/986/data_structures/libE_specs.html

jlnav:
Nope, unfortunately. Although, if we had sub-classes or dictionaries for these common groups, e.g. `DirectorySpecs`, `TCPConfig`, `ResourceSpecs`, etc,  they'd be grouped in the docs. Then such an API may resemble:

```
from libensemble.specs import LibeSpecs, ResourceSpecs

libE_specs = LibeSpecs(
   comms = "local",
   nworkers = 16,
   resources = ResourceSpecs(
      num_resource_sets = 32,
      zero_resource_workers = [1]
   )
)

jmlarson1:
Let's list `comms` as optional before pulling in.

shuds13:
Makes mpi comms optional.

jmlarson1 (approved):

PR #1003, opened 2023-05-09
Refactoring the Ensemble class somewhat

jlnav:
Also starting to _de-emphasize_ the libE function in favor of `Ensemble.run()`

Moving/reorganizing some running-libE content into api.py

Also making Ensemble a normal class. It being a dataclass made autoclass'ing it somewhat ugly.

We're also assuming from here on that specs we're passing into the Ensemble class will be cast to their class counterparts.


jlnav:
rename api.py to ensemble.py

shuds13 (approved):

PR #1017, opened 2023-06-07
Create parent directories of `workflow_dir`

AngelFP:
When using a `workflow_dir`, libEnsemble fails if the `'workflow_dir_path'` has a parent directory that does not exist. This is the case, for example, if `libE_specs['workflow_dir_path'] = './my_workflows/workflow_1'` and the `my_workflows` folder has not yet been created.

This PR makes sure that the parent folders are also created.


jlnav:
Looks good to me, thanks! I'll be adding an additional test soon to validate this behavior will continue to work

Issue #1015, opened 2023-06-06
Warn on nested MPI


jlnav:
MPIExecutor + MPI comms + Open MPI

jlnav:
Closed by #1025

PR #1025, opened 2023-06-14
Warning on nested MPI

jlnav:
Addresses #1015 .

### DIFF
--- a/libensemble/libE.py
+++ b/libensemble/libE.py
@@ -124,6 +124,7 @@ from libensemble.comms.comms import QCommProcess, Timeout
 from libensemble.comms.logs import manager_logging_config
 from libensemble.comms.tcp_mgr import ClientQCommManager, ServerQCommManager
 from libensemble.executors.executor import Executor
+from libensemble.executors.mpi_executor import MPIExecutor
 from libensemble.history import History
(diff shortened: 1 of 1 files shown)


shuds13:
Is it feasible to do this when we have detected OpenMPI. 

That said there could be other OpenMPI based implementations, but usually I've found everything else works okay. 

Comment could say MPICH/MPICH-derived as many implementations like Intel MPI are based on MPICH. Better probably, turn it around and say nesting will not work if using Open MPI.

libensemble/libE.py
```diff
@@ -386,6 +387,13 @@ def libE_mpi_manager(mpi_comm, sim_specs, gen_specs, exit_criteria, persis_info,
     """Manager routine runs on rank 0."""
     from libensemble.comms.mpi import MainMPIComm
 
+    if Executor.executor is not None:
+        if isinstance(Executor.executor, MPIExecutor) and libE_specs["comms"] == "mpi":
```
shuds13, line 391: As inside libE_mpi_manager - I don't think need to test for comms being "mpi" here.

shuds13: Also move below logger set up (which is just below).

shuds13 (approved):
I have updated to only give the message when using Open MPI. We should squash merge this.

PR #1041, opened 2023-07-12
Feature/allow ensemble dir overwrite

jlnav:
User can set ``libE_specs["allow_dir_overwrite"]`` to prevent intentional errors and warnings on attempting to reuse an ensemble directory.

Keep in mind "overwriting" a sim_dir with a new one really means deleting the old and creating the new.

Another consequence of this approach and the current sim/gen-dir naming scheme is if subsequent runs overwrite the same ensemble directory, but different workers each time process the same sim_id, then both sim_dirs for that sim_id will be retained.


shuds13:
Name of option?
Maybe `reuse_ensemble_dir`

We should remove appending of worker ID to each sim_dir so they don't get re-run under different worker and produce another dir.

It is not "overwriting" a sim_dir, but overwriting input files.

shuds13 (commented):
Suggestions:

- Modify top comment on PR (remove "allow_dir_overwrite")
- In ensemble dir exists message - add use `libE_specs["reuse_ensemble_dir"] = True ` to allow reuse of dir.
- created dirs like test_workflow0_nworkers3_comms-local/ need to be cleaned in run_scripts.sh

PR #1049, opened 2023-07-25
Refactor/Rename funcX to Globus Compute

jlnav:
With these updates I discovered that the assignment of resource attributes to the Executor isn't propagated to remote Executor instances, obviously. Pure-python should work in the meantime


shuds13 (approved):

jmlarson1:
I note that the RTD build failed on this branch.

PR #1050, opened 2023-08-01
Add gpu_env_fallback platform attribute

shuds13:
For when setting GPUs to slots in a sim function, without using the MPI executor. Find correct setting for platform.

- [x] Add gpu_env_fallback attribute (used if no env var given to  `set_env_to_gpus` function.
- [x] Add unit tests
- [x] Check shows up in docs

### DIFF
--- a/libensemble/tests/unit_tests/test_platform.py
+++ b/libensemble/tests/unit_tests/test_platform.py
@@ -17,6 +17,7 @@ summit_spec = {
     "gpu_setting_type": "option_gpus_per_task",
     "gpu_setting_name": "-g",
     "scheduler_match_slots": False,
+    "gpu_env_fallback": None,
 }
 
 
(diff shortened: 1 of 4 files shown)


jlnav (approved):

libensemble/tests/unit_tests/test_platform.py
```diff
@@ -17,6 +17,7 @@
     "gpu_setting_type": "option_gpus_per_task",
     "gpu_setting_name": "-g",
     "scheduler_match_slots": False,
+    "gpu_env_fallback": None,
```
shuds13, line 20: This should not be necessary, will default to None. I actually don't know if CUDA_VISIBLE_DEVICES would work on summit under these circumstances, but I can't currently test it.

PR #1051, opened 2023-08-02
Refactor/oo more tests

jlnav:


jmlarson1 (approved):

libensemble/tests/regression_tests/test_GPU_variable_resources.py
```diff
@@ -25,80 +25,68 @@
...
+        out=[("f", float)],
+        user={"dry_run": False},
+    )
+    gpu_test.gen_specs = GenSpecs(
+        gen_f=gen_f,
+        persis_in=["f", "x", "sim_id"],
+        out=[("num_procs", int), ("num_gpus", int), ("x", float, 2)],
+        user={
+            "initial_batch_size": gpu_test.nworkers - 1,
+            "max_procs": (gpu_test.nworkers - 1) // 2,  # Any sim created can req. 1 worker up to max
```
shuds13, line 72: This test needs updating following #1058, I can do this if it helps.

Update this line and add second loop.

shuds13: Whoops - I meant test_GPU_variable_resources_multi_task.py, But I think this line should be changed here also.

shuds13:
I think adding the parse_args option to ensemble would be good, as otherwise, the knowing of workers is a bit magical.

libensemble/tests/regression_tests/test_evaluate_existing_sample.py
```diff
@@ -16,48 +16,33 @@
...
-    H, persis_info, flag = libE(
-        sim_specs, gen_specs, exit_criteria, alloc_specs=alloc_specs, libE_specs=libE_specs, H0=H0
-    )
-
-    if is_manager:
-        assert len(H) == len(H0)
-        assert np.array_equal(H0["x"], H["x"])
-        assert np.all(H["sim_ended"])
+    sampling = Ensemble()
+    sampling.H0 = H0
```
shuds13, line 37: I wonder if we should stick to function calls, matter of style. Or should we make this an option to Ensemble()?

shuds13: ` sampling = Ensemble(H0=H0)`

jlnav: Both ways of setting H0 (and other attributes) are valid:

`sampling = Ensemble(H0=H0)`
or:
```
sampling = Ensemble()
sampling.H0 = H0
```

PR #1052, opened 2023-08-07
Pad calc_dirs with zeros

jlnav:
~~Funnily enough to do this, we have to send exit_criteria to the workers.~~


shuds13:
My understanding of this was something like:

libE_specs["simid_field_width"] = 5.

```
ensemble/
      sim00001
      sim00002
      sim00003`
```

shuds13 (commented):
Maybe a unit test.

docs/data_structures/libE_specs.rst
```diff
@@ -81,6 +81,10 @@ the ``LibeSpecs`` class. When provided as a Python class, options are validated
                     Whether to allow overwrites and access to previous ensemble and workflow directories in subsequent runs.
                     ``False`` by default to protect results.
 
+                "calc_dir_id_width" [int] = ``4``:
+                    The width of the ID component of a calculation directory name. Up to this many additional
+                    zeros are padded to the sim/gen ID.
```
shuds13, line 86: could say "the numerical ID component' to clarify this sentence.

and remove the 'many' as it may not be many. Maybe-
                   
"Leading zeros are padded to the sim/gen ID."

shuds13:
and update the top comment here.

shuds13:
I ran

python test_GPU_variable_resources.py --comms local --nworkers 4

and it gave me a width of four, without adding the option.

shuds13:
This works well, and adds digits on overflow as needed.

I think we just need to check we agree on the option name "calc_dir_id_width".

PR #1053, opened 2023-08-07
swap defaults to False of kill_canceled_sims and safe_mode

jlnav:


shuds13 (commented):
I am going to try optimizing kill_cancelled_sims. Then we can discuss whether to make the breaking change of defaulting to False. I'm hoping it can stay True.

shuds13 (approved):
Squash merge.

PR #1054, opened 2023-08-07
Add mpi_runner_type submit option to executor

shuds13:
The supplied MPI runner will be used for this submit only.

This is useful in combination with the environment script option, when a different MPI runner may be needed for a particular submit.

The option can be given as a string specifying the MPI runner (e.g. "mpich", "openmpi", "srun", "jsrun", "aprun"), or as a dictionary with the same options as `custom_info` when setting up the executor e.g : `{"mpi_runner": "openmpi", "runner_name": "myrunner"}`

Provides support for  optimas-org/optimas#108
- [x] Add functionality test
- [x] Discuss/agree name of argument (currently `mpi_runner_type`). Could be `mpi_runner`, `env_mpi`, `mpi_config` etc...
- [x] Check if this branch MPIExecutor works standalone @jlnav 

Another change is that when using an environment script, the generated script is logged to ensemble.log rather than the mpirun line itself (which is in the script). This could be either way - any feedback welcome.

Ideally, the mpi_runner_type option would match custom_info in executor __init__ function in name and accepted input.

### DIFF
--- a/libensemble/executors/mpi_executor.py
+++ b/libensemble/executors/mpi_executor.py
@@ -15,7 +15,7 @@ resource manager when submitting tasks.
 import logging
 import os
 import time
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import libensemble.utils.launcher as launcher
 from libensemble.executors.executor import Executor, ExecutorException, Task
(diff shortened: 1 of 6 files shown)


jlnav:
second tentative consensus for "mpi_runner_type" ?

jmlarson1 (approved):

shuds13:
Think I'm going to keep `mpi_runner_type`

PR #1057, opened 2023-08-14
Refactor/localopt specific imports

jlnav:
Verified locally works with openmpi


libensemble/gen_funcs/aposmm_localopt_support.py
```diff
@@ -202,6 +195,8 @@ def run_local_nlopt(user_specs, comm_queue, x0, f0, child_can_read, parent_can_r
     Runs an NLopt local optimization run starting at ``x0``, governed by the
     parameters in ``user_specs``.
     """
+    import nlopt
```
jmlarson1, line 198: Can we ensure that moving this import to so late doesn't cause issues when trying to run with (say) 64 workers?

jmlarson1: There was no issue in my tests.

shuds13:
I don't see that the breaking change is necessary. The old rc options would still work - just putting imports in functions and removing the defaults. Either way, we also need a test, and to document when a global import should not be used.

I have a test that uses OpenMPI with the executor, which I can add, so this can be run in the CI. 

Also, mpi4py is working on my laptop, with or without global import, is it failing for you? @jlnav

shuds13 (changes requested):
see comment

PR #1064, opened 2023-08-21
Refactor/petsc openmpi fix

shuds13:
The catch-22 is that when running with mpi4py you must have PETSc import globally, and when running with openMPI (and using the executor) it must NOT import globally, making a one size fits all solution non-trivial. 

The default imports are removed, and rc options are retained. The imports are also added to the functions. If already globally imported (by the rc option), these will not do anything.

- [x] Need to make sure OpenMPI is running on CI for this to be tested.
- [x] Document when to use rc lines, and when not to (OpenMPI/Executor)
- [x] Make sure this page up to date https://libensemble.readthedocs.io/en/develop/examples/aposmm.html
- [x] Remove signal unit test in test_aaa_comms.py

Note: The reason OpenMPI must not import globally (when using executor), is that PETSc imports mpi4py, and if that is done globally, it creates a nested comms, which openMPI does not support.

Note2: As the imports are in the functions, an alternative to the rc line for global import is to just import the package in the user calling script.

- [x] Waiting on #1065


jmlarson1 (approved):

shuds13:
Unit tests for openmpi are failing because MPI being imported - anywhere in unit test dir - does not matter which test.

- It is imported in test_executor.py but as mpi4py.rc.initialize = False - it seems to be okay (this test file works by itself).
- It is in test_manager_main - but I dont think needs to be.
- It also seems to get imported in test_ensemble.py - though not explicitly in that file
  - wondering if when the mpi4py import is removed  in #1065 - if this will go away?
  - otherwise have to put in different dirs.

shuds13:
I dont like that automatic tools dictate lines that go together have been separated. I would prefer the "petsc" line was right under the gen_funcs line, as it originally was.

```python
import libensemble.gen_funcs

# Import libEnsemble items for this test
from libensemble.libE import libE
from libensemble.sim_funcs.six_hump_camel import six_hump_camel as sim_f

libensemble.gen_funcs.rc.aposmm_optimizers = "petsc"
```

Issue #1033, opened 2023-06-23
Support concurrent/futures interface at top level


shuds13:
Allows user to run any generator and use libEnsemble to dispatch simulations

PR #1080, opened 2023-09-11
Experimental/concurrent futures ensemble

jlnav:
Addresses #1033 

Returns a future corresponding to an `Ensemble.run() / libE()` on a single thread.

In a given script, any generator class/process can return points, then slot them into a np array, then we can pass them in a H0 as part of a pregenerated-work workflow


jlnav:
This is *probably* not the implementation we'd want to follow-up on. Closing for now.

PR #1086, opened 2023-09-15
Replace final_fields with final send

jmlarson1:


shuds13:
`final_send` should work to inform the gen, but recall that the option `libE_specs["use_persis_return_gen"]` has to be set also for the generator return to be appended to H. This was presumably to prevent repeated values getting appended.

jmlarson1:
We had discussed removing `final_fields` as an argument and instead returning the normal (e.g. `persis_in`) fields. Is there a reason to want both `final_fields` and `final_send`?

shuds13:
Only that it would break code for user(s) of final_fields. I'm open to removing it.

jmlarson1:
I would vote to remove `final_fields` as an argument. We could error if it's given, or just return all of `persis_in` which would (presumably) include what was in `final_fields`. Perhaps the safest approach would be to return `persis_in` if `final_fields` isn't specified.

jlnav:
I really like the idea anyway of returning persis_in all the time

I was thinking that either for this release or after 1.0, if we decide we want to remove/change a feature, we start doing a more "softwarey" approach. We keep it as-is for that release, but also print/log a DeprecationWarning stating info like:

- This method/option is deprecated
- We'll be removing it next release
- Do the following to port/prepare

PR #1094, opened 2023-09-22
Fixes/into v1.0 release

jlnav:


shuds13 (approved):

shuds13 (commented):
I wouldn't personally capitalise each word of subsections.

shuds13 (changes requested):
I would not capitlize every word of subsections.

shuds13 (approved):

PR #1104, opened 2023-09-26
Fixing dependabot target branch

jmlarson1:


jlnav:
I hope dependabot will reopen its PRs into develop, unless we can change the targets ourselves?

shuds13 (approved):

PR #1112, opened 2023-09-28
Trying different openmpi configs

jmlarson1:


jmlarson1:
We see that 3.9, 3.10, and 3.11 all work with openmpi. We'll keep 3.11

Issue #1132, opened 2023-10-16
Update models to Pydantic 2


jlnav:

jlnav:
Closed by #1135

PR #1135, opened 2023-10-17
Refactor/ support both Pydantic 1 and 2

jlnav:
Addresses #1132 

- [x] Let globus-compute know about possibly updating for pydantic2
- [x] Let Balsam know about possibly updating for pydantic2
- [x] Refactor to support both Pydantic1 and Pydantic2


shuds13 (commented):
Seems to work well with Pydantic 2. Are you going to support 1 and 2 through a import test? How much effort is it?

libensemble/tests/functionality_tests/test_mpi_gpu_settings.py
```diff
@@ -147,10 +147,10 @@
 
         # check having only cores_per_node
         if run_set == "jsrun":
-            del libE_specs["platform_specs"].logical_cores_per_node
+            libE_specs["platform_specs"].logical_cores_per_node = None
```
shuds13, line 150: I hope this doesn't need to be set. Doesn't it get the default in the next call to libE (or .run())

libensemble/tests/functionality_tests/test_mpi_gpu_settings.py
```diff
@@ -60,6 +61,7 @@
 # from libensemble import logger
 # logger.set_level("DEBUG")  # For testing the test
 
+warnings.filterwarnings("ignore", category=UserWarning)
```
shuds13, line 64: What warning do we get here?

shuds13 (approved):

Issue #1103, opened 2023-09-26
New `libe_specs` for saving history file


AngelFP:
There are a couple of features that would be very useful regarding how the history is saved.

1. Saving the history file when the run is completed. Currently it is saved every `k_sims` or `k_gens`, but the last few iterations are missing if, for example, `sim_max` is not a multiple of `k_sims`. This could be implemented by forcing a save on [_final_receive_and_kill](https://github.com/Libensemble/libensemble/blob/52e45af00b217c39f107b2ca5354d7dfdfdff256/libensemble/manager.py#L517). Currently, a workaround is to use this [function](https://github.com/Libensemble/libensemble/blob/52e45af00b217c39f107b2ca5354d7dfdfdff256/libensemble/tools/tools.py#L79C8-L79C8) from the tools, but the name of the file follows a different format and leads to always having 2 history files per run (plus a .picke file).
2. Choose whether the date information should be added to the name of the history file. Having this enabled by default results in many history files being created in optimas when a single exploration is run in several steps or resumed (https://github.com/optimas-org/optimas/pull/118). This is because the `_save_every_k` method is not able to delete the old files, which were created in previous calls to `libE`.
3. As a "nice to have", it would be convenient to have some control on the name of the history file.

Below is an example of how these options could be controlled by the user.
 
```python
libe_specs['save_history_on_completion'] = True
libe_specs['save_history_with_date'] = False   # To save as `libE_history_after_sim_{}.npy`
libe_specs['history_file_prefix'] = 'exploration_history'  # Would allow us to customize it to  `exploration_history_after_sim_{}.npy` in optimas.
```

Let me know what you think.

jlnav:
Hey @AngelFP !

We got around to internally discussing the defaults in #1141 . Just noting here we decided the following:

- `save_H_on_completion` will be False by default, but if a `_save_every_k` option is also set, will be flipped on to True.
- persis_info will also be dumped when save_H_on_completion is set.

AngelFP:
Hi @jlnav, thanks for the update. The first one looks like a good default to me. With respect to the second point, I'm not sure. The setting says `save_H`, so the fact that it also saves persis_info might not be clear to the user. For the particular case of optimas, we don't need the persis_info file, so I would in principle avoid saving it to simplify the output.

Just being a bit nitpicky, it's great to have these options already :D

PR #1141, opened 2023-10-23
Feature / Request History options

jlnav:
Addresses #1103 

New options plus their defaults:

```
    save_H_on_completion: Optional[bool] = False
    """ Save state of ``H`` to file upon completing a workflow."""

    save_H_with_date: Optional[bool] = False
    """ ``H`` filename contains date and timestamp."""

    H_file_prefix: Optional[str] = "libE_history"
    """ Prefix for ``H`` filename."""
```

### DIFF
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -26,27 +26,28 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
 
                 **comms** [str] = ``"mpi"``:
                     Manager/Worker communications mode: ``'mpi'``, ``'local'``, or ``'tcp'``.
+
                 **nworkers** [int]:
                     Number of worker processes in ``"local"`` or ``"tcp"``.
+
                 **mpi_comm** [MPI communicator] = ``MPI.COMM_WORLD``:
                     libEnsemble MPI communicator.
+
                 **dry_run** [bool] = ``False``:
                     Whether libEnsemble should immediately exit after validating all inputs.
+
                 **abort_on_exception** [bool] = ``True``:
                     In MPI mode, whether to call ``MPI_ABORT`` on an exception.
                     If ``False``, an exception will be raised by the manager.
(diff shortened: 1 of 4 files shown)


AngelFP (commented):
Thanks for this @jlnav!
I have given it a try and it works very well. There's only an additional comment I have, which is that the file saved on completion has a different name than those saved every k steps. Ideally this wouldn't be the case, because then we always have two history files after an optimas run (i.e., `exploration_history_after_sim_20.npy` and `exploration_history_after_completion.npy`).

One possible solution that I see would be a final call to
```python
if self.libE_specs.get("save_every_k_sims"):
    self._save_every_k_sims()
if self.libE_specs.get("save_every_k_gens"):
    self._save_every_k_gens()
```
if `"save_H_on_completion"` is enabled. This would probably require a `force=True` argument in `_save_every_k` to make sure that the final history is saved.

jlnav:
> Thanks for this @jlnav! I have given it a try and it works very well. There's only an additional comment I have, which is that the file saved on completion has a different name than those saved every k steps. Ideally this wouldn't be the case, because then we always have two history files after an optimas run (e.g., `exploration_history_after_sim_20.npy` and `exploration_history_after_completion.npy`).
> 
> One possible solution that I see would be a final call to
> 
> ```python
> if self.libE_specs.get("save_every_k_sims"):
>     self._save_every_k_sims()
> if self.libE_specs.get("save_every_k_gens"):
>     self._save_every_k_gens()
> ```
> 
> if `"save_H_on_completion"` is enabled. This would probably require a `force=True` argument in `_save_every_k` to make sure that the final history is saved.

Sounds good, thanks! I can make that change right away.

AngelFP (approved):
Works great now :D

shuds13:
If save_H_on_completion defaults to True, what stops you from getting two .npy files at the end, when user is dumping their file already?

shuds13 (approved):

PR #1143, opened 2023-10-25
Python 3.12

jlnav:


jlnav:
Noting that with a transition to codecov, we can use a newer coverage, which may resolve the python 3.12 warnings

PR #1156, opened 2023-11-06
(Re)implement threading comms

AngelFP:
As discussed in past meetings, we are currently trying to support running Optimas on Jupyter notebooks to allow for a more interactive workflow. However, when testing this a couple of days ago I realized that this is currently not possible due to the use of `multiprocessing` for the local communications. Basically, `multiprocessing` and IPython/Jupyter don't play well together (especially on Windows and Mac, which use `spawn`) and I couldn't get optimas to run without significant workarounds.

The ideal solution would be that libEnsemble supported a `threading` local mode. I gave this a try and also found out that there was an early implementation of this in the `main` branch (was removed in https://github.com/Libensemble/libensemble/pull/1123). I updated the `QCommThread` class to be more similar to `QCommProcess` and exposed a new `local_threading` option in `libe_specs['comms']`. Using this option enables optimas to run on a notebook without any modifications or workarounds by the user.

There are still some issues in the current PR. For example, I had to set `log_comm=False` in the threading backend to prevent an infinite "loop" of logs that keeps filling up the outbox. I suspect this arises somehow because now the workers share the same memory. The shared memory might also lead to issues with how the resources are currently assigned to workers.

Having this threading backend would be great, and the suggested implementation already solves the issues that we need to (urgently) address in optimas. Do you think it is possible to fully support this mode in libEnsemble?

As a bonus, I think this would enable an alternative solution to https://github.com/optimas-org/optimas/pull/134, where we could define an option such as `libe_specs["run_persis_gen_on_thread"]=True` to run the generator on a thread (that is, it would be able to share the memory with the manager).

### DIFF
--- a/libensemble/comms/comms.py
+++ b/libensemble/comms/comms.py
@@ -221,8 +221,17 @@ class QCommThread(Comm):
         return self._result
 
     def terminate(self, timeout=None):
-        """Terminate the thread."""
-        raise Timeout()
+        """Terminate the thread.
+        
+        A thread can't really be killed from the outside. Ideally the `main`
+        function would make periodic checks to some variable that determines
+        whether the function should continue. This is not implemented, so
+        it is currently no possible to terminate the thread when calling
+        this method.
+        """
+        self.thread.join(timeout=timeout)
+        if self.running:
+            raise Timeout()
 
     @property
     def running(self):
--- a/libensemble/comms/comms.py
+++ b/libensemble/comms/comms.py
@@ -222,7 +222,7 @@ class QCommThread(Comm):
 
     def terminate(self, timeout=None):
         """Terminate the thread.
-        
+
         A thread can't really be killed from the outside. Ideally the `main`
         function would make periodic checks to some variable that determines
         whether the function should continue. This is not implemented, so
--- a/libensemble/comms/comms.py
+++ b/libensemble/comms/comms.py
@@ -28,6 +28,7 @@ from abc import ABC, abstractmethod
 
 # from multiprocessing import Process, Queue, Value, Lock
 from multiprocessing import Process, Queue
+from threading import Thread
 from time import time
 from traceback import format_exc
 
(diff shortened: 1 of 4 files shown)


jlnav:
Hey @AngelFP ! We'd be happy to formally support threading comms, and its especially great that the implementation was easy since its restoring code we recently removed :)  Thanks!

Either you or we can do this, but much of `test_comms.py` in `libensemble/tests/unit_tests` also needs to be restored.

AngelFP:
Hmm, I had a look at the old tests just now, and they require these classes that are not used anywhere else. Would it be better to define new tests for `QCommThread` only? Like those in `tests/test_aaa_comms.py`

jlnav:
> Hmm, I had a look at the old tests just now, and they require these classes that are not used anywhere else (and which I didn't have to restore). Would it be better to define new tests for `QCommThread` only that mirror those of `QCommProcess`? Like those in `tests/test_aaa_comms.py`

Sounds good to me! (although looking at those tests they aren't as substantial as I remembered, but maybe that's okay since the comms are tested everywhere else)

shuds13:
We can support, and see how well this works, but I would say use with caution. Threads share an environment, including current working directory. Setting enviornment variables like CUDA_VISIBLE_DEVICES will be shared, unless we move that into a sub-process. These kind of issues are why we did not use the threading interface in the past.

AngelFP:
> We can support, and see how well this works, but I would say use with caution. Threads share an environment, including current working directory. Setting enviornment variables like CUDA_VISIBLE_DEVICES will be shared, unless we move that into a sub-process. These kind of issues are why we did not use the threading interface in the past.

Makes sense. The idea from the optimas side would be to enable threading only for light function evaluations that do not require the allocation of resources (num_procs or num_gpus) nor the creation of simulation directories. So, basically, only for user-defined functions evaluated with the `FunctionEvaluator`, not the `TemplateEvaluator` that we use for running heavy simulations.

shuds13 (approved):

PR #1162, opened 2023-11-08
Fix two bugs in Ensemble class

jlnav:


shuds13:
We want to make sure this is tested. Our regression tests run with comms/nworkers on command line, but if parse_args=False, we can check it gets the libE_specs values and sets ensemble.nworkers.That said, as ensemble.libE_specs is set after, I presume the values here would overwrite command line values anyway.

shuds13 (approved):

PR #1168, opened 2023-11-12
Fix Ensemble instantiation and lines beneath if name == main in tutorial

jlnav:


jlnav:
https://github.com/Libensemble/libensemble/actions/runs/6843217648

shuds13 (commented):
Gen uses Input/Output instead of H H_out - diff to ALL our other examples.

Naming of files eg.

Create a new Python file named simulator.py. Write the following:
then gives:
examples/tutorials/simple_sine/tutorial_sim.py

Why does simple_sine  tutorial call it tutorial_sim.py - its already in the dir name.

When I try to run I get:

> 
> (libe_tut_test) shudson@shuds:tut_test$  python calling_script.py
> Traceback (most recent call last):
>   File "/home/shudson/tut_test/calling_script.py", line 2, in <module>
>     from libensemble import Ensemble, LibeSpecs, SimSpecs, GenSpecs, ExitCriteria
> ImportError: cannot import name 'LibeSpecs' from 'libensemble' (/home/shudson/miniconda3/envs/libe_tut_test/lib/python3.12/site-packages/libensemble/__init__.py)

In 1d sampling we have:
from libensemble.specs import ExitCriteria, GenSpecs, LibeSpecs, SimSpecs

Did this run for you?

shuds13:
I would also recommend that at least the files under examples/tutorial/simple_sine exactly match the tutorial, that's what I did for the forces tutorial. Both in contents and in the name of the files.

shuds13:
We need to decide this.
```
from tutorial_gen import gen_random_sample
from tutorial_sim import sim_find_sine
```

I feel like this should point to the scripts you just created. 

I see that above works without having done the gen/sims step, but presumably the user may want to play around the the scripts they created, so this should point to their generator.py and simulator.py. You could preface it by saying create a new directory and create scripts inside it.

shuds13 (approved):

PR #1169, opened 2023-11-14
Feature/default local

shuds13:
Aim to default to local when an MPI runner is not used to run libEnsemble script.

This is not a definitive way to know if a process has been run using an MPI runner but the following could be used.

1. Check for environment variables such as PMI_SIZE and OMPI_COMM_WORLD_SIZE
2. We could assume if `nworkers` is given in libE_specs, that it defaults to local.

This should include support of parse_args when --nworkers is provided on the command line but not comms.

Add unit (and possibly functionality) test/s for different configurations - including when parse_args=False and when parse_args=True and comms is not supplied.


shuds13:
For now I have removed the function `detect_comms_env` as it seems sufficient to default to **local** comms if `nworkers` is set and **mpi** comms otherwise. 

In the class interface, if `ensemble.libE_specs` is first set without comms or nworkers - it will default comms to mpi. If `ensemble.libE_specs.nworkers` is then set, this will not change to local as it does not know if the comms was set by the user or default. However, as mpi comms seems incompatible with nworkers I have made the setting of both an error.

shuds13:
I have also assumed local comms in parse_args if only nworkers is set, and allowed the short option `-n`. So if using parse_args:

`python my_script.py -n 4`

Will run with local comms and 4 workers.

shuds13:
Okay, its all changed again.

The function `detect_comms_env` is back, but called only when there is no comms specified and `nworkers` is set.

In this scenario
- if an MPI environment is detected (with more than one process), it will keep **mpi** comms (as the user has probably run with an MPI runner, but will issue a warning that `nworkers` will be ignored).
-  If an MPI environment is not detected it will default to **local** comms. 

So the following code should use MPI comms is run with an MPI runner, but local comms otherwise.

```python

sampling = Ensemble()
sampling.libE_specs = LibeSpecs(nworkers=3)
```

If `nworkers` is not set, it will default to **mpi** comms, without detection, as the detection only works for MPI runners that set the recognized environment variables, and there is no reason to think they are trying to use `local` comms.

jlnav (approved):

PR #1175, opened 2023-11-22
Feature/decorators simplify examples

jlnav:
Tentative refactoring of a handful of _mostly_ functionality tests based on decorated example functions


jmlarson1:
@jlnav This greatly simplifies our many tests, which is nice. But I'm concerned it hides the necessary input/output relationships between the sims and gens. It is nice to have the decorators available, but we should think about how much we want to use them.

jlnav:
> @jlnav This greatly simplifies our many tests, which is nice. But I'm concerned it hides the necessary input/output relationships between the sims and gens. It is nice to have the decorators available, but we should think about how much we want to use them.

Good point. We know for a given sim and gen that they're compatible if their outputs and inputs correspond. 

I just realized though, suppose a user tries to combine two user functions that are currently incompatible, their ins and outs don't match up. Changing the ins and outs within sim_specs and gen_specs _doesn't solve their incompatibility_! In fact, adjusting the ins/outs may cause the specs to pass validation, but errors would occur within the user function anyway.

I'll probably undo the regression test changes, since those are more user-examples.

libensemble/specs.py
```diff
@@ -528,7 +530,9 @@ def persistent_uniform(_, persis_info, gen_specs, libE_info):
 
     def decorator(func):
         setattr(func, "persis_in", fields)
-        func.__doc__ = f"\n    **Persistent Input Fields:** ``{func.persis_inputs}``\n" + func.__doc__
+        if not func.__doc__:
+            func.__doc__ = ""
```
jmlarson1, line 534: Is there no test using `persis_in`? (Codecov says it's not covered)

libensemble/tests/scaling_tests/forces/forces_adv/run_libe_forces.py
```diff
@@ -63,7 +63,6 @@
 # State the generating function, its arguments, output, and necessary parameters.
 gen_specs = {
     "gen_f": gen_f,  # Generator function
-    "in": [],  # Generator input
```
shuds13, line 66: Is removing this line necessary. I would rather not change it.

libensemble/tests/scaling_tests/forces/forces_simple/run_libe_forces.py
```diff
@@ -41,7 +41,6 @@
 
     ensemble.gen_specs = GenSpecs(
         gen_f=gen_f,
-        inputs=[],  # No input when start persistent generator
```
shuds13, line 44: as above

libensemble/specs.py
jlnav: These uncovered lines refer to user functions without docstrings, something that thankfully we have very few of! I'll see what I can do

shuds13 (approved):

PR #1177, opened 2023-11-24
Allow overwriting history file on completion

AngelFP:
This PR allows `_save_every_k` to overwrite the last saved history file on completion (when using the `"save_H_on_completion"` and/or `"save_every_k"` options). Otherwise, when using the `"final_gen_send"` option, there were cases in which "gen_informed" was still `False` in the saved file, while being `True` in the history array.


jlnav (approved):
Looks good to me, thank you very much!

shuds13 (approved):

PR #1184, opened 2023-12-04
Testing / misc fixes

jlnav:
Most notably, don't need to add TESTSUITE lines to the tutorial-code tests since they'll be run with comms=local, nworkers=4 defaults.


jlnav:
Looks like we have a 3.12 fail because matplotlib, used in the tutorials, is using a deprecated time method

jmlarson1 (approved):

PR #1213, opened 2024-01-04
Giving the option of updating gpCAM using a grid of points

jmlarson1:


shuds13:
@jmlarson1 
Should `new_inds` returned from `_find_eligible_points` (line 144) be appended rather than overwritten?

Possible optimizations for `_find_eligible_points`?
- `sorted_X` is not really used - so can remove
- `sorted_indices` and `sorted_D` look like they are invariant within a batch so calc outside while loop (line 143).
- Not sure the deep copies are necessary?

I can make these changes if they make sense - 2x in regression test.

Using this gen (without my optimizations) in libE/WarpX runs, I'm immediately getting extremely small variances (essentially zero) -  so something is not right, don't know if error is in gen or something else in my setup though.

shuds13 (approved):

PR #1215, opened 2024-01-05
Briefly document threads comms, plus warn that sim_dirs aren't supported with threads

jlnav:


shuds13:
I would rather not document threads at the moment. If we do, we need to be clear that resources will also not work.

Issue #1166, opened 2023-11-09
Support running generator on the manager


shuds13:
This may be accomplished by one of the following approaches:

1. Run the generator serially (in-place) on the manager
2. Run the generator on a thread

Advantages for common use-case of a single generator:

- Simplifies determination of the number of workers
- Pass arguments by reference
- Reduces a communication

jlnav:
Closed by recent release!

PR #1216, opened 2024-01-09
Refactor/user function handling modules +  Manager can run additional worker on thread

jlnav:
Addresses #1166 

# User - Manager runs local worker on thread - Gen Worker 0

The manager can start an additional worker _thread_ with the worker ID 0. 

This worker is, via `W["gen_worker"] = True`, and filtering options, made to run gens only.

Tentative interface: `libE_specs["gen_on_manager"] = True`

# Internal - Refactor runners.py

Largely refactors utils/runners.py so additional "Runners" for user functions can be more easily developed/extended.

GlobusComputeRunner and ThreadRunner inherit from Runner.

Runner inspects the contents of the input sim/gen specs and instantiates the correct subclass on the left-hand-side:

```
>>> sim_specs = {... , "globus_compute_endpoint": "1234}
>>> runner = Runner(sim_specs)
>>> runner
<libensemble.utils.runners.GlobusComputeRunner object at 0x1490ec190>
>>> runner = Runner({"sim_f": print})
>>> runner
<libensemble.utils.runners.Runner object at 0x104e024d0>
```

# Internal -  _WorkerIndexer class for indexing into worker arrays

# Internal - Workers can be assigned gen_worker.

### DIFF
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -30,12 +30,9 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
                 **nworkers** [int]:
                     Number of worker processes in ``"local"``, ``"threads"``, or ``"tcp"``.
 
-                **manager_runs_additional_worker** [bool] = False
-                    Manager process launches an additional threaded Worker 0.
-                    This worker can access/modify user objects by reference.
-
                 **gen_on_manager** Optional[bool] = False
-                    Enable ``manager_runs_additional_worker`` and reserve that worker for a single generator.
+                    Instructs Manager process to run generator functions.
+                    This generator function can access/modify user objects by reference.
 
                 **mpi_comm** [MPI communicator] = ``MPI.COMM_WORLD``:
                     libEnsemble MPI communicator.
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -55,6 +55,10 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
                 **disable_log_files** [bool] = ``False``:
                     Disable ``ensemble.log`` and ``libE_stats.txt`` log files.
 
+                **gen_workers** [list of ints]:
+                    List of workers that should only run generators. All other workers will only
+                    run simulator functions.
+
         .. tab-item:: Directories
 
             .. tab-set::
--- a/docs/data_structures/libE_specs.rst
+++ b/docs/data_structures/libE_specs.rst
@@ -30,8 +30,12 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
                 **nworkers** [int]:
                     Number of worker processes in ``"local"``, ``"threads"``, or ``"tcp"``.
 
-                **manager_runs_additional_worker** [int] = False
-                    Manager process can launch an additional threaded worker
+                **manager_runs_additional_worker** [bool] = False
+                    Manager process launches an additional threaded Worker 0.
+                    This worker can access/modify user objects by reference.
+
+                **gen_on_manager** Optional[bool] = False
+                    Enable ``manager_runs_additional_worker`` and reserve that worker for a single generator.
 
                 **mpi_comm** [MPI communicator] = ``MPI.COMM_WORLD``:
                     libEnsemble MPI communicator.
--- a/libensemble/alloc_funcs/only_one_gen_alloc.py
+++ b/libensemble/alloc_funcs/only_one_gen_alloc.py
@@ -21,22 +21,26 @@ def ensure_one_active_gen(W, H, sim_specs, gen_specs, alloc_specs, persis_info, 
     gen_flag = True
     gen_in = gen_specs.get("in", [])
 
-    for wid in support.avail_worker_ids():
-        persis_info = support.skip_canceled_points(H, persis_info)
+    persis_info = support.skip_canceled_points(H, persis_info)
 
-        if persis_info["next_to_give"] < len(H):
+    if persis_info["next_to_give"] < len(H):
+        for wid in support.avail_worker_ids(gen_workers=False):
             try:
                 Work[wid] = support.sim_work(wid, H, sim_specs["in"], [persis_info["next_to_give"]], [])
             except InsufficientFreeResources:
                 break
             persis_info["next_to_give"] += 1
+            if persis_info["next_to_give"] >= len(H):
+                break
 
-        elif not support.test_any_gen() and gen_flag:
+    elif not support.test_any_gen() and gen_flag:
+        # Give gen work
+        return_rows = range(len(H)) if gen_in else []
+        for wid in support.avail_worker_ids(gen_workers=True):
+
             if not support.all_sim_ended(H):
                 break
 
-            # Give gen work
-            return_rows = range(len(H)) if gen_in else []
             try:
                 Work[wid] = support.gen_work(wid, gen_in, return_rows, persis_info.get(wid))
             except InsufficientFreeResources:
--- a/libensemble/alloc_funcs/only_one_gen_alloc.py
+++ b/libensemble/alloc_funcs/only_one_gen_alloc.py
@@ -35,16 +35,16 @@ def ensure_one_active_gen(W, H, sim_specs, gen_specs, alloc_specs, persis_info, 
     elif not support.test_any_gen() and gen_flag:
         # Give gen work
         return_rows = range(len(H)) if gen_in else []
-        for wid in support.avail_worker_ids(gen_workers=True):
+        wid = support.avail_worker_ids(gen_workers=True)[0]
 
-            if not support.all_sim_ended(H):
-                break
+        if not support.all_sim_ended(H):
+            return Work, persis_info
 
-            try:
-                Work[wid] = support.gen_work(wid, gen_in, return_rows, persis_info.get(wid))
-            except InsufficientFreeResources:
-                break
-            gen_flag = False
-            persis_info["total_gen_calls"] += 1
+        try:
+            Work[wid] = support.gen_work(wid, gen_in, return_rows, persis_info.get(wid))
+        except InsufficientFreeResources:
+            return Work, persis_info
+        gen_flag = False
+        persis_info["total_gen_calls"] += 1
 
     return Work, persis_info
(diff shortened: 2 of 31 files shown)


libensemble/manager.py
```diff
@@ -154,6 +155,48 @@ def filter_nans(array: npt.NDArray) -> npt.NDArray:
 """
 
 
+class _Worker:
+    """Wrapper class for Worker array and worker comms"""
+
+    def __init__(self, W: npt.NDArray, wid: int, wcomms: list = []):
+        self.__dict__["_W"] = W
```
shuds13, line 162: How different from self._W = W

shuds13:
From a quick look, I don't understand the number of calls to the _Worker constructor. Dont we just need to create one wrapper class? This could use the __getitem__ method to set indexing. e.g.

class _Worker:
    def __init__(self, W, manager_as_worker=False):
        self.data = data
        self.manager_as_worker = manager_as_worker

    def __getitem__(self, key):
        if self.manager_as_worker:
            return self.data[key]
        else:
            return self.data[key - 1]

jlnav:
Yeah, I don't like the number of calls myself to the _Worker constructor. I was leaning towards something like in manager.run:

```python
for w in Work:
    if self._sim_max_given():
        break
  w = _Worker(
  self._check_work_order(Work[w], w)
  ...
```

But avoided it at the time since I didn't want to rewrite too much logic also dealing with passing worker IDs around e.g.:

```python
            if w in self.persis_pending:
                self.persis_pending.remove(w)
                worker.active = 0
            self._freeup_resources(w)
```

So you're imagining the interface being something like `self.worker[w].active = 0` then? Yeah, the dynamic indexing would still work and so would the attribute setting/getting, plus the same class wouldn't have to be made multiple times.

shuds13:
You'll see in #1142 I made a change in worker_resources to allow assigning resources to the manger.

jlnav:
TODO on additional branch:

- [ ] Ensure resources are correctly allocated, especially considering "resources" objects being identical for the manager/worker_0 as they're in the same interpreter
- [ ] Check dumped_every_k persis_info matches that maintained by worker 0 within a persistent gen

shuds13:
Its debatable but I'm inclined to make use of extra worker transparent to user. Maybe "gen_on_thread" or something. Not "manager_runs_additional_worker".

libensemble/manager.py
```diff
@@ -208,6 +230,34 @@ def __init__(
             (1, "stop_val", self.term_test_stop_val),
         ]
 
+        if self.libE_specs.get("manager_runs_additional_worker", False):
```
shuds13, line 233: I would think should be an "else" here.

jlnav: I'm not sure what would go in "else", though. We do extra setup for an additional worker, but there's nothing alternative to do otherwise.

shuds13: The lines that are always done above
        else:
            self.W = np.zeros(len(self.wcomms), dtype=Manager.worker_dtype)
            self.W["worker_id"] = np.arange(len(self.wcomms)) + 1

libensemble/comms/comms.py
```diff
@@ -146,10 +146,11 @@ def mail_flag(self):
 
 
 class QCommLocal(Comm):
-    def __init__(self, main, nworkers, *args, **kwargs):
+    def __init__(self, main, *args, **kwargs):
         self._result = None
         self._exception = None
         self._done = False
+        self._ufunc = kwargs.get("ufunc", False)
```
shuds13, line 153: self._ufunc appears not to be used currently

libensemble/comms/comms.py
```diff
@@ -208,10 +209,13 @@ def result(self, timeout=None):
         return self._result
 
     @staticmethod
-    def _qcomm_main(comm, main, *args, **kwargs):
+    def _qcomm_main(comm, main, *fargs, **kwargs):
```
shuds13, line 212: I would probably keep as *args by convention

libensemble/comms/comms.py
```diff
@@ -208,10 +209,13 @@ def result(self, timeout=None):
...
         """Main routine -- handles return values and exceptions."""
         try:
-            _result = main(comm, *args, **kwargs)
+            if not kwargs.get("ufunc"):
```
shuds13, line 215: Call something like user_func to not confuse with ufunc for universal functions.

libensemble/manager.py
```diff
@@ -597,7 +652,7 @@ def _alloc_work(self, H: npt.NDArray, persis_info: dict) -> dict:
 
         alloc_f = self.alloc_specs["alloc_f"]
         output = alloc_f(
-            self.W,
+            self.W.iterable,
```
shuds13, line 658: I'm uneasy about this. Should we not use self.W in the same way in alloc funcs (inc. in alloc support). Esp. as there is potentially invisible errors if you end up with W[-1]

jlnav: Good point. I'll refactor more of alloc_support

libensemble/tools/alloc_support.py
```diff
@@ -87,29 +87,22 @@ def assign_resources(self, rsets_req, use_gpus=None, user_params=[]):
             rset_team = self.sched.assign_resources(rsets_req, use_gpus, user_params)
         return rset_team
 
-    def avail_worker_ids(self, persistent=None, active_recv=False, zero_resource_workers=None):
+    def avail_worker_ids(self, persistent=False, active_recv=False, zero_resource_workers=None, worker_type=None):
```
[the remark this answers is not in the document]
jlnav: When we can filter for specific types of workers, gen workers or sim workers, then their persistent state doesn't also need to specify persistent_sim or persistent_gen. There's no situation where a gen worker should have a persistent_sim state, so its best that "persistent" describes the worker as a bool

libensemble/manager.py
jlnav: One complication I just realized is that that'll expose the _WorkerIndexer wrapper to the user. `W` as traditionally within the alloc_f would no longer be be a NumPy array unless we pass in the iterable as we do here.

jlnav: Ah, but if we don't explicitly use W outside of persistent_support functions, then we can wrap it with an indexer inside PersistentSupport

libensemble/tools/alloc_support.py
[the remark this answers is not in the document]
shuds13: None means give me a list of all workers, persistent or not.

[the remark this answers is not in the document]
shuds13: Perhaps we don't use it. But we should agree on it.

libensemble/manager.py
shuds13: I'm leaning towards we pass self.W, even if its breaking. I will review the alloc funcs.

libensemble/manager.py
```diff
@@ -158,13 +159,36 @@ class Manager:
 
     worker_dtype = [
         ("worker_id", int),
+        ("worker_type", int),
         ("active", int),
-        ("persis_state", int),
-        ("active_recv", int),
+        ("persistent", bool),
```
shuds13, line 164: Are these changes necessary?

shuds13 (changes requested):
What I would do:
 - Make sure W changes are consistent (either 3 booleans + worker type or as it was)
 - persistent (default changed to False rather than None - how does this affect things.)
 - Make gen_worker field for W - mirrors zero_resource_worker field.
 - Pass self.W to alloc and modify as necessary (I think).
 - testing
   - test giving gen diff resources.

libensemble/tools/alloc_support.py
```diff
@@ -87,24 +87,19 @@ def assign_resources(self, rsets_req, use_gpus=None, user_params=[]):
             rset_team = self.sched.assign_resources(rsets_req, use_gpus, user_params)
         return rset_team
 
-    def avail_worker_ids(self, persistent=None, active_recv=False, zero_resource_workers=None):
+    def avail_worker_ids(self, persistent=None, active_recv=False, zero_resource_workers=None, gen_workers=False):
```
shuds13, line 90: Shouldn't gen_workers default to None, like zero_resource_workers?

libensemble/tools/alloc_support.py
```diff
@@ -87,24 +87,19 @@ def assign_resources(self, rsets_req, use_gpus=None, user_params=[]):
...
         """Returns available workers as a list of IDs, filtered by the given options.
 
         :param persistent: (Optional) Int. Only return workers with given ``persis_state`` (1=sim, 2=gen).
         :param active_recv: (Optional) Boolean. Only return workers with given active_recv state.
         :param zero_resource_workers: (Optional) Boolean. Only return workers that require no resources.
+        :param gen_workers: (Optional) Boolean. If True, return gen-only workers and manager's ID.
```
shuds13, line 96: why return managers ID separatly is that necessary

jlnav:
> I think the only issue now, is thread-lock not getting cleaned up between libE() calls.

This is fixed. I'll merge when approved.

docs/data_structures/libE_specs.rst
```diff
@@ -28,7 +28,11 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
                     Manager/Worker communications mode: ``'mpi'``, ``'local'``, or ``'tcp'``.
 
                 **nworkers** [int]:
-                    Number of worker processes in ``"local"`` or ``"tcp"``.
+                    Number of worker processes in ``"local"``, ``"threads"``, or ``"tcp"``.
```
jmlarson1, line 31: but not `mpi`?

libensemble/alloc_funcs/only_one_gen_alloc.py
```diff
@@ -21,22 +21,25 @@ def ensure_one_active_gen(W, H, sim_specs, gen_specs, alloc_specs, persis_info,
     gen_flag = True
     gen_in = gen_specs.get("in", [])
 
-    for wid in support.avail_worker_ids():
-        persis_info = support.skip_canceled_points(H, persis_info)
-
-        if persis_info["next_to_give"] < len(H):
+    if persis_info["next_to_give"] < len(H):
+        for wid in support.avail_worker_ids(gen_workers=False):
+            persis_info = support.skip_canceled_points(H, persis_info)
```
jmlarson1, line 26: Do we only want to skip canceled points if the above two conditions are satisfied?

docs/data_structures/libE_specs.rst
jlnav: I don't believe the nworkers option has ever influenced mpi comms, only `mpirun -n N` makes a difference

libensemble/alloc_funcs/only_one_gen_alloc.py
jlnav: The one condition?

shuds13:
@jlnav 

- In start_only_persistent why not gen_worker=False on the sim
- In inverse bayes - should H["gen_worker"] by gen_workers? (if so we need a test).
- What about fast_alloc_and_pausing and give_pregenerated_work?

shuds13 (approved):

PR #1236, opened 2024-02-07
Experimental/class gens ask tell

jlnav:
Addresses #1240 

Approach where the _**worker**_ itself, outside of calling any user-function, interacts with an ask/gell gen.

1. Initialize the Generator object on the user-side, so they can **parameterize however they'd like**.
2. Insert into gen_specs.
3. On the worker, evaluate if our gen_f has a .ask. If so, setup the corresponding Runner.
4. Upon "calling" the gen, if the corresponding libE_info indicates the gen is _not_ persistent, simply call the gen's .ask.
5. Otherwise start a familiar persistent loop that asks the gen, sends data outward, and tells responses to the gen.
6. Upon a stop-tag being received, cleanup via .final_tell.

The included functionality test works locally :)


libensemble/generators.py
```diff
@@ -0,0 +1,227 @@
...
+        array["sim_id"] = np.arange(self.sim_id_index, self.sim_id_index + len(array))
+        self.sim_id_index += len(array)
+        return array
+
+    def initial_ask(self, num_points: int = 0, *args) -> npt.NDArray:
+        return super().initial_ask(num_points, args)[0]
+
+    def ask(self, num_points: int = 0) -> (npt.NDArray, Optional[npt.NDArray]):
+        _, self.last_ask = self.outbox.get()
+        points = self._add_sim_ids(self.last_ask["calc_out"])
```
shuds13, line 216: Not clear to me why expect points then cancels. In Surmise gen_f main loop, there is a send_recv which will send points and then wait (for a tell). The next comm is a cancel (if there is one).

shuds13:
I'm wondering how the interfacer gens could be used in something like Optimas, as we have different interfacers for APOSMM/Surmise, with their own pattern of ask/tell calls?

PR #1264, opened 2024-03-14
Bugfix/ensemble libe specs attrs passthrough

jlnav:


shuds13:
Needs fix to work with pydantic v1

jmlarson1 (approved):

Issue #1268, opened 2024-03-20
Release v1.3.0


shuds13:
Target: May 01 2024

Before release:
- [x] Check documentation. Is gen_on_manager and change to default comms documented sufficiently.
- [x] Polaris build (conda module disappeared - still in Polaris docs) - see if need to update Polaris instructions (tested using cray-python module) Now also tested with conda module - need first use: module use /soft/modulefiles

Features:

- [x] Support generator running on the manager on thread. #1216/#1290
- [x] Default to local comms when num_workers is supplied and no MPI runner detected .#1169
- [x] Parse args defaults to local comms when nworkers is set on command line (via --nworkers or -n). #1169

Possible:

- [ ] Improve coverage of core libE methods (if reasonably easy to do so). Files to consider
  - [ ] [libE.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2FlibE.py)
    - [x] `on_abort`, `on_cleanup` method
    - [ ] TCP worker-side content.
  - [ ] Should all of [worker.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fworker.py) be covered?
  - [ ] [comms.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fcomms%2Fcomms.py)
  - [x] [executor.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fexecutors%2Fexecutor.py)
  - [x] [mpi_runner.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fexecutors%2Fmpi_runner.py)
    - [x] including `_adjust_procs`
  - [ ] [mpi_resources.py](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fresources%2Fmpi_resources.py) exception handling
  - [ ] [genfunc init](https://app.codecov.io/github/Libensemble/libensemble/blob/develop_coverage/libensemble%2Fgen_funcs%2F__init__.py) where I think much might be removable.
[...]

PR #1298, opened 2024-04-25
Release/v 1.3.0

shuds13:
Ref #1268


jlnav (approved):

docs/data_structures/libE_specs.rst
```diff
@@ -26,9 +26,15 @@ libEnsemble is primarily customized by setting options within a ``LibeSpecs`` cl
 
                 **comms** [str] = ``"mpi"``:
                     Manager/Worker communications mode: ``'mpi'``, ``'local'``, or ``'tcp'``.
+                    If ``nworkers`` is specified, then ``local`` comms will be used unless a
+                    parallel MPI environment is detected.
```
jmlarson1, line 30: Do we need the word "parallel" here?

shuds13: It checks there is more than one process. On some platforms you can srun into a bash session on a compute node and you are technically in an MPI environment. We don't want that to count.

PR #1306, opened 2024-05-03
Tentatively *deprecate* use_persis_return_sim/gen, zero_resource_workers, AllocSpecs.out

jlnav:
We've worried for a long time that it's impossible to remove things without inconveniencing/warning users manually. For features/capabilities we'd like to remove or dramatically change, we ought to raise DeprecationWarnings for those features, warning of exactly when to expect changes.

I think nowadays users seem to be more likely to request new features, changes, or cleanups at the cost of complete backward compatibility. Plus, thankfully, we're not deleting old releases; I wouldn't mind releasing patches to old releases if requested, similarly to how Python 3.9, 3.10, etc. still get patches.

So based on a handful of conversations, I'll start by proposing the deprecation of those features listed in the title.


jlnav:
zero_resource_workers removed in another branch. Other features' removals will be addressed separately.

PR #1307, opened 2024-05-09
ASK/TELL DEVELOP

jlnav:
See checklist in https://github.com/Libensemble/libensemble/pull/1307#pullrequestreview-2210671527


libensemble/tests/unit_tests/test_persistent_aposmm.py
```diff
@@ -168,8 +168,88 @@ def test_standalone_persistent_aposmm_combined_func():
...
+            "ftol_abs": 1e-6,
+            "dist_to_bound_multiple": 0.5,
+            "max_active_runs": 6,
+            "lb": np.array([-3, -2]),
+            "ub": np.array([3, 2]),
+        },
+    }
+
+    my_APOSMM = APOSMM(gen_specs)
+    my_APOSMM.setup()
```
shuds13, line 206: Does this need to be separate to constructor?

jlnav: Perhaps you've seen already, but `setup()` sets attributes that can't be pickled, but need to be done anyway. So I believe a separate `generator.setup()` needs to exist.

But perhaps for outside-of-libE purposes, to save a line, this could resemble: `my_APOSMM = APOSMM(gen_specs, setup=True)`.

shuds13:
Currently this has two existing user functions refactored to ask/tell, and in a way that is a breaking change. I don't think this is what we want. So, for now, we could supply these as duplicates (keeping the original), or remove them. Alternative would be to refactor all appropriate gens to ask/tell, but that would hold this up.

jlnav:
> @jlnav
> 
> Currently this has two existing user functions refactored to ask/tell, and it is a breaking change for codes using those gens. I don't think this is what we want. So, for now, we could supply these as duplicates (keeping the original), or remove them. Alternative would be to refactor all appropriate gens to ask/tell, but that would hold this up, so I think it may be better to supply these two as a duplicate for now.

Sounds good. Maybe we could raise DeprecationWarnings?

> Also, would it be better for these ask/tell gens (rand sample and gpCAM) to use AskTellGenRunner and not use the wrapper gen_f?

I prefer the AskTellGenRunner myself (I _did_ develop it), but if all the gens move to my runner, then what would you want to do with your wrapper?

shuds13:
I think there is also some opportunity for inheritence with the gpCAM class. And gp_cam_simple is more complicated than gp_cam_asktell. So, if anything the latter would be the base class, and renamed to avoid confusion.

libensemble/libE.py
```diff
@@ -441,6 +441,24 @@ def libE_mpi_worker(libE_comm, sim_specs, gen_specs, libE_specs):
 # ==================== Local version ===============================
 
 
+def _retrieve_generator(gen_specs):
```
shuds13, line 444: I don't know why this is needed. If gen is on a thread, should not need to be pickled.

shuds13: I've removed on my branch for now.

jlnav:
Checking random sample when running without Optimas:


```
>>> import numpy as np
>>> normal = np.load("persistent_aposmm_nlopt_history_length=2005_evals=2000_workers=4.npy")
>>> asktell = np.load("persistent_aposmm_nlopt_asktell_history_length=2002_evals=2000_workers=3.npy")
>>> all([i in asktell["x"][:100] for i in normal["x"][:100]])
True
```

jlnav:
With Optimas, currently looks like random sample doesn't match, even when trying to account for seed

jlnav:
Precision fixed in the Optimas example: https://github.com/optimas-org/optimas/pull/209/commits/08a835bb2d4efa30b76b9d86acf071c7d544310d

shuds13:
We may need to update gpCAM gens for latest gpCAM release. Make sure any changes made in gen_f is reflected here before pulling in.

- [ ] Check gpCAM updates

libensemble/gen_classes/gpCAM.py
```diff
@@ -0,0 +1,150 @@
...
+            self.y_new = np.atleast_2d(calc_in["f"]).T
+            nan_indices = [i for i, fval in enumerate(self.y_new) if np.isnan(fval)]
+            self.x_new = np.delete(self.x_new, nan_indices, axis=0)
+            self.y_new = np.delete(self.y_new, nan_indices, axis=0)
+
+            self.all_x = np.vstack((self.all_x, self.x_new))
+            self.all_y = np.vstack((self.all_y, self.y_new))
+
+            if self.my_gp is None:
+                self.my_gp = GP(self.all_x, self.all_y, noise_variances=self.noise * np.ones(len(self.all_y)))
```
shuds13, line 92: I need to update for new gpCAM interface

libensemble/gen_classes/sampling.py
```diff
@@ -0,0 +1,48 @@
...
+class UniformSample(LibensembleGenerator):
+    """
+    This generator returns ``gen_specs["initial_batch_size"]`` uniformly
+    sampled points the first time it is called. Afterwards, it returns the
+    number of points given. This can be used in either a batch or asynchronous
+    mode by adjusting the allocation function.
+    """
+
+    def __init__(self, _, persis_info, gen_specs, libE_info=None) -> list:
+        # self.H = H
```
shuds13, line 21: Remove I guess.

libensemble/gen_classes/gpCAM.py
```diff
@@ -0,0 +1,150 @@
...
+        self.ub = np.array(user_specs["ub"])
+        self.n = len(self.lb)  # dimension
+        assert isinstance(self.n, int), "Dimension must be an integer"
+        assert isinstance(self.lb, np.ndarray), "lb must be a numpy array"
+        assert isinstance(self.ub, np.ndarray), "ub must be a numpy array"
+        self.all_x = np.empty((0, self.n))
+        self.all_y = np.empty((0, 1))
+        np.random.seed(0)
+
+    def __init__(self, H, persis_info, gen_specs, libE_info=None):
```
shuds13, line 54: I will put above _initialize_gpcAM

shuds13: and make  _initialize_gpcAM  _initialize_gpCAM

libensemble/gen_classes/gpCAM.py
```diff
@@ -0,0 +1,150 @@
...
+        self.lb = np.array(user_specs["lb"])
+        self.ub = np.array(user_specs["ub"])
+        self.n = len(self.lb)  # dimension
+        assert isinstance(self.n, int), "Dimension must be an integer"
+        assert isinstance(self.lb, np.ndarray), "lb must be a numpy array"
+        assert isinstance(self.ub, np.ndarray), "ub must be a numpy array"
+        self.all_x = np.empty((0, self.n))
+        self.all_y = np.empty((0, 1))
+        np.random.seed(0)
+
```
shuds13, line 53: We need to decide __init__ interface. 
We questioned before where we keep the same interface - which mirrors the current gen_f, or to rearrange, as H is often not given (basically an H0). So it could be gen_specs first. I'm leaning towards keeping the original ordering as it mirrors our user functions, but this should be discussed. 

We could also

shuds13 (changes requested):
If we pull this in, are we treating this as provisional, without documenting gen_classes, so interface (e.g. `__init__` option ordering) is mutable. Or are we ready to say a user can use the gen_classes interface for their own gens if desired? This questions applies to the numpy interface (`ask_np`/`tell_np`). As for the outer API, we certainly have not finalized that, and it should somehow be noted.

E.g.,
```# This feature is in beta and its interface may change in future releases```

Note, that currently the two implemented classes (rearranged to ask/tell) are duplicates of gen_f, but in the long term we will not want to have duplicates.

Discussions:
- [x] Agree current status of gen_classes (internal or public).
- [x] If provisional - mark somehow in code
- [x] Agree wrapper paradigm (outer v inner interface)
- [x] Agree naming ask_np / tell_np
  - still not sure
  ```python
    ask_numpy()
   ```
   is better than 
   ```python
     @convert_dict_to_numpy
     ask()
   ```
  It seems we need separate function names, and the output of an ask() would be ambiguous.

- [x] Do we allow passing a class for ask/tell (if a libE generator) so initialize internally.
 - this is currently supported with the wrapper method but not internal ask/tell method
 - I think we should probably not support both wrapper and ask/tell approaches.
- [x] Agree num_points for ask policy (initial and thereafter):
 - Always empty ask and generator has to decide (most similar to now but what about external gens?)
 - intercept:
   -  init: `gen_specs['user']` fields `initial_batch_size` or `gen_batch_size` and else use num workers
   - after: intercept num points coming in for tell and ask for that many
   - Problem with intercepting `gen_specs['user']`  is names for batch size may be different.
 - Add fixed gen_specs options `gen_specs['initial_batch_size']`
   - and maybe something like `gen_specs['return_policy']` (or better name)
     - fixed batch size or generate num points got back or let generator decide points - empty ask()
     - (at some point may replace alloc_specs `async_return` if we want most user not to worry about allocs)?
 

- [x] Data conversion - how treat multi-dimensional fields.
 - currently makes a numpy array (as a dictionary value).
 
- [x] Agree gpCAM class names (e.g., GP_CAM_Covar) - and match gen_func names (original functions).
- [ ] Naming convention for tests - we have:
  - test_gpCAM_class.py
  - test_asktell_surmise.py
  - test_persistent_surmise_killsims_asktell.py
  - test_persistent_aposmm_nlopt_asktell.py
  - test_sampling_asktell_gen.py

- [x] Do our gens all support passing in intiail points? And if so, in standard format?

Interfacer:
- [ ] Does it work with executor (consideration of things added to executor in worker.py - including first-tier comm object)
- [x] Does it work when there is no executor
- [ ] For user function should we insert comm into libE_info["comm"] inside qcomm_main
- [x] Works with processes and threads - or picked one (and named right in generators.py)
- [x] Works with fork and spawn (for spawn be clear what in calling script needs to be inside main block).
- [x] Sometimes get hang at end of runs using the interfacer - esp. if no final_tell, is there something not getting shut-down? Check this works.
- [x] Need to deal with final_tell when using gen (e.g. optimas)
- [x] Need to sepaarate final_tell as required by standard - tell and extract data function.
- [ ] Constructor wrapper approach - demonstrate alternatives in notebook

Todo:
- [x] SH - update gpCAM to reflect current interface.
- [x] Change/rename original gpCAM to match naming in the class.
- [x] Need to run 'extra' tests for gpCAM tests to get run. Check passes.
- [x] Check correct no. points produced each call of gen (usually number sent back).

Check works:
- [x] Running libE ask/tell generator - e.g. gpCAM through Optimas (it will go via ask/tell wrappers).
- [x] Use libEnsemble ask/tell generator standalone (outside of libE)
 - must libE be installed to use generators?
- [x] Running libE with our ask_numpy/tell_numpy functions.
- [x] Running libE with ask/tell function (standardised format).

File: libensemble/gen_classes/gpCAM.py (hunk 6 above)
Comment (jlnav) [reply libensemble/gen_classes/gpCAM.py:53]: Fair enough. My opinion/intuition is a user is more likely to prefer either "classical" gens (e.g. Jeff) or ask/tell gens (e.g. other CAMPA folks). With these gens' interfaces and users being so different, I don't think an arguably simpler rearrangement of the input parameters is too confusing.

Similarly to how some people prefer numpy or pandas; they do similar things, but their interfaces being different isn't a point of contention.

I'd also lean towards if someone were to initialize some object, like a gen, themselves, they'd prefer their specifications be provided as early and clearly as possible:

```
my_gen = Generator(param=1,
     option="two",
)
```

vs.

```
my_gen = Generator(None,
     {},
     {"param": 1, "option": "two"},
     {},
)
```

File: libensemble/generators.py [hunk 7]
```diff
@@ -0,0 +1,293 @@
...
+from libensemble.comms.comms import QComm, QCommThread
+from libensemble.executors import Executor
+from libensemble.message_numbers import EVAL_GEN_TAG, PERSIS_STOP
+from libensemble.tools import add_unique_random_streams
+from libensemble.utils.misc import list_dicts_to_np, np_to_list_dicts
+
+
+class Generator(ABC):
+    """
+    v 0.7.2.24
```
jmlarson1, line 18: Is this version number something we need?

jmlarson1 (approved):

libensemble/generators.py
```diff
@@ -0,0 +1,300 @@
...
+        optionally retrieve an updated final state of evaluations. This is a separate
+        method to simplify the common pattern of noting internally if a
+        specific tell is the last. This will be called only once.
+        """
+
+
+class LibensembleGenerator(Generator):
+    """Internal implementation of Generator interface for use with libEnsemble, or for those who
+    prefer numpy arrays. ``ask/tell`` methods communicate lists of dictionaries, like the standard.
+    ``ask_np/tell_np`` methods communicate numpy arrays containing the same data.
```
shuds13, line 92: Check docstring names match functions (ask_numpy?)

libensemble/generators.py
```diff
@@ -0,0 +1,300 @@
...
+        self.tell_numpy(list_dicts_to_np(calc_in))
+
+
+class LibensembleGenThreadInterfacer(LibensembleGenerator):
+    """Implement ask/tell for traditionally written libEnsemble persistent generator functions.
+    Still requires a handful of libEnsemble-specific data-structures on initialization.
+    """
+
+    def __init__(
+        self, gen_specs: dict, History: npt.NDArray = [], persis_info: dict = {}, libE_info: dict = {}
```
shuds13, line 118: Here we have a different constructor signature to those functions in gen_classes.

shuds13:
In the full tests: https://github.com/Libensemble/libensemble/actions/runs/10256911231/job/28376940998#step:22:684

```
 ---Test 7: test_asktell_surmise.py starting with local on 4 processes 
/home/runner/miniconda3/envs/condaenv/lib/python3.11/site-packages/pydantic/_internal/_config.py:322: UserWarning: Valid config keys have changed in V2:
* 'orm_mode' has been renamed to 'from_attributes'
  warnings.warn(message, UserWarning)
Traceback (most recent call last):
  File "/home/runner/work/libensemble/libensemble/libensemble/tests/regression_tests/test_asktell_surmise.py", line 90, in <module>
    H_out, _a, _b = borehole(list_dicts_to_np(point), {}, sim_specs, {"H_rows": np.array([point["sim_id"]])})
                             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/libensemble/libensemble/libensemble/utils/misc.py", line 102, in list_dicts_to_np
    first = list_dicts[0]
            ~~~~~~~~~~^^^
KeyError: 0
Error: The operation was canceled.
```

libensemble/utils/runners.py
```diff
@@ -84,3 +97,117 @@ def _result(self, calc_in: npt.NDArray, persis_info: dict, libE_info: dict) -> (
...
+    def _result(self, calc_in: npt.NDArray, persis_info: dict, libE_info: dict) -> (npt.NDArray, dict, Optional[int]):
+        if libE_info.get("persistent"):
+            return self._persistent_result(calc_in, persis_info, libE_info)
+        return self._to_array(self.gen.ask(getattr(self.gen, "batch_size", 0) or libE_info["batch_size"]))
+
+
+class LibensembleGenRunner(AskTellGenRunner):
+    def _get_initial_ask(self, libE_info) -> npt.NDArray:
+        """Get initial batch from generator based on generator type"""
+        H_out = self.gen.ask_numpy(self.initial_batch or libE_info["batch_size"])
```
shuds13, line 173: How are these batch sizes determined. How do we support fixed and/or dynamic sizes (e.g., producing as many points as has been returned)?

libensemble/utils/runners.py
```diff
@@ -84,3 +97,117 @@ def _result(self, calc_in: npt.NDArray, persis_info: dict, libE_info: dict) -> (
...
+    def _get_points_updates(self, batch_size: int) -> (npt.NDArray, npt.NDArray):
+        return self._to_array(self.gen.ask(batch_size)), self._to_array(self.gen.ask_updates())
+
+    def _convert_tell(self, x: npt.NDArray) -> list:
+        self.gen.tell(np_to_list_dicts(x))
+
+    def _loop_over_gen(self, tag, Work):
+        """Interact with ask/tell generator that *does not* contain a background thread"""
+        while tag not in [PERSIS_STOP, STOP_TAG]:
+            batch_size = self.batch or Work["libE_info"]["batch_size"]
```
shuds13, line 130: How are these batch sizes determined. How do we support fixed and/or dynamic sizes (e.g., producing as many points as has been returned)? Same below.

libensemble/utils/runners.py
shuds13: 2. I thought we said you call ask_numpy if exists, else call ask() - or is that dealt with by inheriting from Generator instead of LibensembleGenerator. Is there a test with a Generator in the standardised format using list of dictionary interface? We could add one in gen_classes sampling.

libensemble/generators.py
```diff
@@ -0,0 +1,300 @@
...
+    """
+
+    def __init__(
+        self, gen_specs: dict = {}, History: npt.NDArray = [], persis_info: dict = {}, libE_info: dict = {}, **kwargs
+    ) -> None:
+        from libensemble.gen_funcs.persistent_aposmm import aposmm
+
+        gen_specs["gen_f"] = aposmm
+        if len(kwargs) > 0:
+            gen_specs["user"] = kwargs
```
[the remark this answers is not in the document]
jlnav: Ah, its a weird looking choice that made the following interface possible on the Optimas side, and possibly the libE side too if I refactor the aposmm_asktell regression test:

```
aposmm = APOSMM(
    initial_sample_size=100,
    localopt_method="LN_BOBYQA",
    sample_points=np.round(known_minima, 1),
    rk_const=0.5 * ((gamma(1 + (n / 2)) * 5) ** (1 / n)) / sqrt(pi),
    xtol_abs=1e-6,
    ftol_abs=1e-6,
    dist_to_bound_multiple=0.5,
    max_active_runs=4,
    lb=np.array([var_1.lower_bound, var_2.lower_bound]),
    ub=np.array([var_1.upper_bound, var_2.upper_bound]),
)
```

libensemble/generators.py
```diff
@@ -0,0 +1,300 @@
...
+
+    @abstractmethod
+    def ask_numpy(self, num_points: Optional[int] = 0) -> npt.NDArray:
+        pass
+
+    @abstractmethod
+    def tell_numpy(self, results: npt.NDArray) -> None:
+        pass
+
+    def ask(self, num_points: Optional[int] = 0) -> List[dict]:
```
shuds13, line 103: Should these not be the definitions in Generator?

libensemble/generators.py
```diff
@@ -0,0 +1,300 @@
...
+        """Request the next set of points to evaluate."""
+        return np_to_list_dicts(self.ask_numpy(num_points))
+
+    def tell(self, calc_in: List[dict]) -> None:
```
shuds13, line 107: should calc_in be 'results' - same as in Generator?

libensemble/utils/runners.py
```diff
@@ -84,3 +97,117 @@ def _result(self, calc_in: npt.NDArray, persis_info: dict, libE_info: dict) -> (
...
+            points, updates = self._get_points_updates(batch_size)
+            if updates is not None and len(updates):  # returned "samples" and "updates". can combine if same dtype
+                H_out = np.append(points, updates)
+            else:
+                H_out = points
+            tag, Work, H_in = self.ps.send_recv(H_out)
```
shuds13, line 136: Dont we need a `np_to_list_dicts(H_out)` before this. 

Also, in the list of dictionaries data structure. If we convert a multi-dimensional x field, what type is the dictionary value supposed to be? Is it supposed to be a numpy array. I wonder this because the idea of the simple data structures was in part meant to avoid numpy.

shuds13:
For APOSMM example, H is dumped before final_tell call, and so does not have local_min in dumped H. Also, looks like other fields might be a bit different between dumped H and the one returned from final_tell.

jlnav:
> For APOSMM example, H is dumped before final_tell call, and so does not have local_min in dumped H. Also, looks like other fields might be a bit different between dumped H and the one returned from final_tell.

Yeah; this will be true for Optimas until they implement `final_tell`. Until then Optimas currently has no way of knowing what points APOSMM has internally marked as minima (since `ask_updates` was rejected).

Since this branch does implement `final_tell` _and_ respects APOSMM's updates via marking points in H, **libEnsemble**'s H and APOSMM's H should be perfectly in sync.

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,123 @@
...
+                ("sim_id", int),
+                ("local_min", bool),
+                ("local_pt", bool),
+            ]
+            gen_specs["persis_in"] = ["x", "f", "local_pt", "sim_id", "sim_ended", "x_on_cube", "local_min"]
+        if not persis_info:
+            persis_info = add_unique_random_streams({}, 2, seed=4321)[1]
+        super().__init__(History, persis_info, gen_specs, libE_info, **kwargs)
+        if not self.persis_info.get("nworkers"):
+            self.persis_info["nworkers"] = gen_specs["user"]["max_active_runs"]  # ??????????
```
shuds13, line 37: This is not nworkers. It could be less than nworkers.

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,123 @@
...
+            self._last_ask = super().ask_numpy(num_points)
+            if self._last_ask[
+                "local_min"
+            ].any():  # filter out local minima rows, but they're cached in self.all_local_minima
+                min_idxs = self._last_ask["local_min"]
+                self.all_local_minima.append(self._last_ask[min_idxs])
+                self._last_ask = self._last_ask[~min_idxs]
+        if num_points > 0:  # we've been asked for a selection of the last ask
+            results = np.copy(
+                self._last_ask[self._ask_idx : self._ask_idx + num_points]
```
shuds13, line 82: When we are in this situation (perhaps in error) where APOSMM does not give us enough points (e.g. the current ask is for 6, but APOSMM has only supplied one) then this gets filled up with empty points. This should be handled somehow. If it should never arise, then as a error condition.

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,123 @@
...
+    def _array_size(self):
+        """Output array size must match either initial sample or N points to evaluate in parallel."""
+        user = self.gen_specs["user"]
+        return user["initial_sample_size"] if not self._told_initial_sample else user["max_active_runs"]
+
+    @property
+    def _enough_initial_sample(self):
+        """We're typically happy with at least 90% of the initial sample, or we've already told the initial sample"""
+        return (
+            self._n_buffd_results >= self.gen_specs["user"]["initial_sample_size"] - 10
```
shuds13, line 64: this line looks very dubious. Why is 10 a hardcoded value?

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,123 @@
...
+        ):  # Optimas prefers to give back chunks of initial_sample. So we buffer them
+            self._tell_buf = np.zeros(self._array_size, dtype=self.gen_specs["out"] + [("f", float)])
+
+        if not self._enough_initial_sample:
+            self._slot_in_data(np.copy(results))
+            self._n_buffd_results += len(results)
+        self._n_total_results += len(results)
+
+        if not self._told_initial_sample and self._enough_initial_sample:
+            self._tell_buf = self._tell_buf[self._tell_buf["sim_id"] != 0]
```
shuds13, line 110: Why getting rid of sim_id 0?

libensemble/gen_classes/aposmm.py
jlnav: Right. I'll find a better way to specify nworkers to aposmm

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,123 @@
...
+        self.all_local_minima = []
+        self._ask_idx = 0
+        self._last_ask = None
+        self._tell_buf = None
+        self._n_buffd_results = 0
+        self._n_total_results = 0
+        self._told_initial_sample = False
+
+    def _slot_in_data(self, results):
+        """Slot in libE_calc_in and trial data into corresponding array fields. *Initial sample only!!*"""
```
shuds13, line 47: this is what you want:
`self._tell_buf[self._n_buffd_results:self._n_buffd_results + len(results)] = results`

libensemble/gen_classes/aposmm.py
shuds13: If its not used until the main loop, can it be captured, or atleast checked later?

libensemble/generators.py
```diff
@@ -0,0 +1,206 @@
...
+    def ask_numpy(self, num_points: Optional[int] = 0) -> npt.NDArray:
+        """Request the next set of points to evaluate, as a NumPy array."""
+
+    @abstractmethod
+    def tell_numpy(self, results: npt.NDArray) -> None:
+        """Send the results, as a NumPy array, of evaluations to the generator."""
+
+    def ask(self, num_points: Optional[int] = 0) -> List[dict]:
+        """Request the next set of points to evaluate."""
+        return np_to_list_dicts(self.ask_numpy(num_points))
```
shuds13, line 118: I think we will want to convert from numpy types to regular python types. This could be done inside np_to_list_dicts or here e.g.

```python
    @staticmethod
    def convert_np_types(dict_list):
        np_types = (np.float64, np.int64, np.int32, np.float32)
        return [
            {key: (value.item() if isinstance(value, np_types) else value) for key, value in item.items()}
            for item in dict_list
        ]

    def ask(self, num_points: Optional[int] = 0) -> List[dict]:
        """Request the next set of points to evaluate."""
        vals = LibensembleGenerator.convert_np_types(np_to_list_dicts(self.ask_numpy(num_points)))
        return vals
```

libensemble/tests/regression_tests/test_asktell_aposmm_nlopt.py
```diff
@@ -0,0 +1,101 @@
...
+        initial_batch_size=10,
+        user={"initial_sample_size": 100},
+    )
+
+    workflow.libE_specs.gen_on_manager = True
+    workflow.add_random_streams()
+
+    H, persis_info, _ = workflow.run()
+
+    # Perform the run
```
shuds13, line 87: Move comment up?

shuds13:
Consequences of removing the gen_f wrapper.
Orchestration,  in terms of how you interface with a model, happens inside a gen_f.
This includes things like where we decide resources for each simulation.
For example, in the Optimus gen_f, it has an option to set `num_procs` or `num_gpus` for each simulation. This could be used with any generator . If we get rid of the gen_f wrapper where does this happen?

libensemble/gen_classes/aposmm.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(
+        self,
+        variables: dict,
+        objectives: dict,
+        History: npt.NDArray = [],
+        persis_info: dict = {},
+        gen_specs: dict = {},
+        libE_info: dict = {},
+        **kwargs,
```
shuds13, line 24: For libEnsemble arguments this is opaque when looking at this routine. Perhaps they should be written out as additional arguments. Then there's the question of whether you keep gen_specs as well.

shuds13 (approved):

PR #1311, opened 2024-05-17
Examples/remove notebook metadata

jmlarson1:
What do you think about removing metadata from these? (If you wanted that metadata, then feel free to close this PR)


jlnav:
Yeah, if I recall right the metadata should be removed so that "our" state isn't cached

shuds13 (approved):

PR #1321, opened 2024-06-04
Feature/reverse ssh nodelist

jlnav:
Run with:

`python any_calling_script.py --worker_file worker_target.yaml`

Starts reverse-ssh connect to specified host, with worker_id range.

Can specify which remote python and which remote working directory


jlnav:
Closing for now, not a priority

PR #1328, opened 2024-06-12
First experiment with creating an ask/tell gen that creates the agreed-upon data-type

jlnav:


libensemble/gen_classes/gpCAM.py
```diff
@@ -113,7 +115,7 @@ def __init__(self, H, persis_info, gen_specs, libE_info=None):
             self.x_for_var = _generate_mesh(self.lb, self.ub, self.num_points)
             self.r_low_init, self.r_high_init = _calculate_grid_distances(self.lb, self.ub, self.num_points)
 
-    def ask(self, n_trials):
+    def ask(self, n_trials) -> list:
```
shuds13, line 118: should type hint specify list of dictionaries?

shuds13:
I was thinking the idea was to create a wrapper?

shuds13:
Of course a wrapper means we need to decide some way to know when it is to be used v native libE interface.  There are ways we could do this in the generator using a decorator somehow. 

Just as example, though this would mean a decorator on the generator ask.

```python

import numpy as np
from functools import wraps

def convert_input(func):
    @wraps(func)
    def wrapper(self, input_data, *args, **kwargs):
        if isinstance(input_data, list) and all(isinstance(item, dict) for item in input_data):
            input_data = self.convert_dict_to_np(input_data)
        return func(self, input_data, *args, **kwargs)
    return wrapper

class Generator:
    def __init__(self):
        pass

    def convert_dict_to_np(self, list_of_dicts):
        dtype = [(key, type(value)) for key, value in list_of_dicts[0].items()]
        structured_array = np.array([tuple(d.values()) for d in list_of_dicts], dtype=dtype)
        return structured_array

    @convert_input
    def gen_function(self, structured_array):
        # Original generator function logic using structured_array
        print(repr(structured_array))

generator = Generator()

# List of dictionaries input
list_of_dicts = [{'a': 1, 'b': 2.0}, {'a': 3, 'b': 4.0}]
generator.gen_function(list_of_dicts)

# Numpy structured array input
structured_array = np.array([(1, 2.0), (3, 4.0)], dtype=[('a', int), ('b', float)])
generator.gen_function(structured_array)

```

There may be other ways to redirect based on incoming data.

shuds13:
You could also apply the decorators dynamically in libensemble. In the example, try removing `@convert_input` from gen_function and do:

```python
setattr(Generator, "gen_function", convert_input(Generator.gen_function))
```

shuds13:
An alternative approach is to make an ask() function that supports the dictionary interface and an internal function, e.g. ask_numpy(), that implements using numpy arrays. So ask() is the data conversion wrapper. That way, libEnsemble can call ask_numpy() directly if it exists in a generator, and otherwise, convert to standard interface (e.g. dictionaries) and call ask(). 

Pros: Avoids fancy decorators that obfuscate.

libensemble/gen_classes/gpCAM.py
```diff
@@ -61,7 +64,13 @@ def __init__(self, H, persis_info, gen_specs, libE_info=None):
         self.my_gp = None
         self.noise = 1e-8  # 1e-12
 
-    def ask(self, n_trials):
+    def ask(self, num_points: Optional[int] = 0) -> List[dict]:
```
shuds13, line 67: I guess these wrappers could go at a higher level, though may reduce visibility to end users. 
Could be base class -> libE class -> generators.

shuds13 (approved):

PR #1332, opened 2024-06-17
No longer using deprecated in1d method

jmlarson1:
https://numpy.org/doc/stable/reference/generated/numpy.in1d.html is being removed in numpy 2.0. It seems as if `isin` works the same.


jmlarson1:
After pulling this in, we should be able to update #1330.

shuds13 (approved):

Issue #1331, opened 2024-06-17
Improve setting of is_manager and nworkers in ensemble class


shuds13:
When parse_args is not used, ensemble class does not set nworkers or is_manager.

This is somewhat non-trivial due to the different order in which things can be set. However, it is particularly problematic that `if ensemble.is_manager` does not work at the end of a script.

I have looked at setting these in the libE_specs setter function when they are set in new_specs, but suprisingly, this seems to contain default values in addition to those explicitly specified. I don't want to assume a default MPI here if it is not set, so maybe we need to duplicate what we do in the validators in `default_comms`.

shuds13:
Fixed by #1336

PR #1336, opened 2024-06-27
Feature/enforce first comms

jlnav:
Please let me know if I'm possibly missing any edge-cases or possibilities for simplification.

Helps address #1331


jlnav:
Currently displaying UserWarnings upon trying to overwrite comms. It would be easy enough to change these to ValueErrors.

jmlarson1 (approved):

shuds13 (changes requested):
Issues:

`test_1d_sampling_no_comms_given.py` is failing on local comms. We need to decide what should happen here.

The above test, and my original test code, has no `sampling.is_manager` True when run with MPI. Notice the print and save file does not happen.

jlnav:
> Issues:
> 
> `test_1d_sampling_no_comms_given.py` is failing on local comms. We need to decide what should happen here.
> 
> The above test, and my original test code, has no `sampling.is_manager` True when run with MPI. Notice the print and save file does not happen.

It's failing because with parse_args=False, it defaults to an empty LibeSpecs() (MPI). Then we try updating the LibeSpecs with nworkers (local), so as decided, it fails.

Perhaps we can remove the test. I think that fail-condition is covered by the `test_ensemble_prevent_comms_overwrite` and `test_local_comms_without_nworkers` unit tests. I'd only keep it for coverage purposes if the execution somehow proceeds _BEYOND_ where the unit tests already test for intentional ValueErrors

shuds13:
- [ ] We also need to check, no examples in docs or anywhere that would not work with this.

shuds13:
> It's failing because with parse_args=False, it defaults to an empty LibeSpecs() (MPI). Then we try updating the LibeSpecs with nworkers (local), so as decided, it fails.

I think the correct behavior is to ignore nworkers when MPI is decided so the test works either way. I can change this.

shuds13:
> Perhaps we can remove the test. I think that fail-condition is covered by the `test_ensemble_prevent_comms_overwrite` and `test_local_comms_without_nworkers` unit tests. I'd only keep it for coverage purposes if the execution somehow proceeds _BEYOND_ where the unit tests already test for intentional ValueErrors

We definitely want a full parallel test on this. We are looking at making sure is_manager and nworkers are correctly set.

shuds13 (approved):
I think this is ready now.

PR #1398, opened 2024-08-09
Fix zero GPU case when GPUs assigned to worker

shuds13:
When GPUs have been assigned to a worker, and num_gpus=0 is sent, it was assigning all available GPUs.
- [x] Check single node
- [x] Check multiple nodes

This PR also changes some scenarios where resources are under-utilized for a submit, such that the given configuration will be fit to the minimum nodes required.

E.g., if a worker has been assigned 2 nodes but gets a run configuration of num_procs=2 and num_gpus=2, and this can be accommodated on one node, then only one node will be used, rather than splitting over two.

### DIFF
--- a/libensemble/executors/mpi_executor.py
+++ b/libensemble/executors/mpi_executor.py
@@ -326,7 +326,7 @@ class MPIExecutor(Executor):
         if not num_procs and not match_procs_to_gpus:
             num_procs = self.gen_nprocs
 
-        if not num_gpus:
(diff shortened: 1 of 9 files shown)


jlnav (approved):

n01r:
I tested this PR with a scaled-down version of my setup, and it fixed the issue. Thanks a bunch, @shuds13! :rocket:

jmlarson1 (approved):

PR #1408, opened 2024-08-15
Fixing broken link

jmlarson1:
When building docs locally, I get an error:

`libensemble/libensemble/resources/platforms.py:docstring of libensemble.resources.platforms.Known_platforms.perlmutter:1: WARNING: py:class reference target not found: libensemble.resources.platforms.Perlmutter [ref.class]`

I was hoping that fixing the broken URL would resolve this. It doesn't.


jlnav:
Need to add `libensemble.resources.platforms.Perlmutter` or whatever to docs/nitpicky

jlnav (approved):

PR #1409, opened 2024-08-19
Feature/flatten asktell dict values

jlnav:
- Packing kwargs into gen_specs["user"] now occurs in `LibensembleGenerator` parent class
- multi-dim values from numpy flattened into e.g. x0, x1 when converted to list-of-dicts. These get packed back up into (x, int, (2,)) in list_dicts_to_np.

### DIFF
--- a/libensemble/gen_classes/sampling.py
+++ b/libensemble/gen_classes/sampling.py
@@ -31,11 +31,10 @@ class UniformSample(SampleBase):
     mode by adjusting the allocation function.
     """
 
-    def __init__(self, _, persis_info, gen_specs, libE_info=None):
-        self.persis_info = persis_info
-        self.gen_specs = gen_specs
-        self.libE_info = libE_info
+    def __init__(self, _=[], persis_info={}, gen_specs={}, libE_info=None, **kwargs):
+        super().__init__(gen_specs, _, persis_info, libE_info, **kwargs)
         self._get_user_params(self.gen_specs["user"])
+        self.gen_specs["out"] = [("x", float, (self.n,))]
 
     def ask_numpy(self, n_trials):
         H_o = np.zeros(n_trials, dtype=self.gen_specs["out"])
@@ -57,11 +56,10 @@ class UniformSampleDicts(Generator):
     mode by adjusting the allocation function.
     """
 
-    def __init__(self, _, persis_info, gen_specs, libE_info=None):
-        self.persis_info = persis_info
-        self.gen_specs = gen_specs
-        self.libE_info = libE_info
+    def __init__(self, _, persis_info, gen_specs, libE_info=None, **kwargs):
+        super().__init__(_, persis_info, gen_specs, libE_info)
         self._get_user_params(self.gen_specs["user"])
+        self.gen_specs["out"] = [("x", float, (self.n,))]
 
     def ask(self, n_trials):
         H_o = []
--- a/libensemble/gen_classes/sampling.py
+++ b/libensemble/gen_classes/sampling.py
@@ -34,7 +34,6 @@ class UniformSample(SampleBase):
     def __init__(self, _=[], persis_info={}, gen_specs={}, libE_info=None, **kwargs):
         super().__init__(_, persis_info, gen_specs, libE_info, **kwargs)
         self._get_user_params(self.gen_specs["user"])
-        self.gen_specs["out"] = [("x", float, (self.n,))]
 
     def ask_numpy(self, n_trials):
         H_o = np.zeros(n_trials, dtype=self.gen_specs["out"])
@@ -60,7 +59,6 @@ class UniformSampleDicts(Generator):
         self.gen_specs = gen_specs
         self.persis_info = persis_info
         self._get_user_params(self.gen_specs["user"])
-        self.gen_specs["out"] = [("x", float, (self.n,))]
 
     def ask(self, n_trials):
         H_o = []
--- a/libensemble/gen_classes/sampling.py
+++ b/libensemble/gen_classes/sampling.py
@@ -57,7 +57,8 @@ class UniformSampleDicts(Generator):
     """
 
     def __init__(self, _, persis_info, gen_specs, libE_info=None, **kwargs):
-        super().__init__(_, persis_info, gen_specs, libE_info)
+        self.gen_specs = gen_specs
+        self.persis_info = persis_info
         self._get_user_params(self.gen_specs["user"])
         self.gen_specs["out"] = [("x", float, (self.n,))]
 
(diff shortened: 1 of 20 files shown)


shuds13 (changes requested):
We can discuss the theory, but the implementation is partial, and would need to be made consistent.

shuds13:
This is the test data I'm using.

```python
dtype = [('a', 'i4'), ('x', 'f4', (3,)), ('z', 'f4', (12,)), ('greeting', 'U10'), ('co2', 'f8')]
H = np.zeros(2, dtype=dtype)
H[0] = (1, [1.1, 2.2, 3.3],[1,2,3,4,5,6,7,8,9,10,11,12], 'hello','1.23')
H[1] = (2, [4.4, 5.5, 6.6],[51,52,53,54,55,56,57,58,59,60,61,62], 'goodbye','2.23')

list_dicts = np_to_list_dicts(H)
npp = list_dicts_to_np(list_dicts)

for field in H.dtype.names:
    print(f"Comparing {field}: {H[field]} {npp[field]}")

    if isinstance(H[field], np.ndarray):
        assert np.array_equal(H[field], npp[field]), f"Mismatch found in field {field}"

    elif isinstance(H[field], str) and isinstance(npp[field], str):
        assert H[field] == npp[field], f"Mismatch found in field {field}"

    elif np.isscalar(H[field]) and np.isscalar(npp[field]):
        assert np.isclose(H[field], npp[field]), f"Mismatch found in field {field}"

    else:
        raise TypeError(f"Unhandled or mismatched types in field {field}: {type(H[field])} vs {type(npp[field])}")
```

jlnav:
Starting to lean towards a `target_dtype` parameter for `list_dicts_to_np`. It _could_ be helpful, but there's also the dilemma where we're not using these conversion functions on generators that specify a dtype in the first place, _our_ generators...

And as we've discussed it would be best not to force users of external gens to specify dtypes...?

libensemble/generators.py
```diff
@@ -90,6 +91,18 @@ class LibensembleGenerator(Generator):
...
 
+    def __init__(
+        self, History: npt.NDArray = [], persis_info: dict = {}, gen_specs: dict = {}, libE_info: dict = {}, **kwargs
+    ):
+        self.gen_specs = gen_specs
+        if len(kwargs) > 0:  # so user can specify gen-specific parameters as kwargs to constructor
+            self.gen_specs["user"] = kwargs
+        if not persis_info:
+            self.persis_info = add_unique_random_streams({}, 4, seed=4321)[1]
+            self.persis_info["nworkers"] = 4
```
shuds13, line 102: whats going on here

jlnav: The kwargs line is similar to https://github.com/Libensemble/libensemble/pull/1307#discussion_r1705983300

for `persis_info`, if libE ask/tell generators maintain the `H, persis_info, gen_specs, libE_info, **kwargs` signature, my thinking is users may prefer not needing to populate/maintain persis_info.

Its the circumstance where 

```python
aposmm = APOSMM(
    ub=1,
    lb=0,
)
```

is probably preferred to:

```python
aposmm = APOSMM(
    None,
    {},
    {
        "user": {"ub": 1, "lb": 0},
    },
    {}
)
```

libensemble/gen_classes/sampling.py
```diff
@@ -57,11 +56,11 @@ class UniformSampleDicts(Generator):
...
     """
 
-    def __init__(self, _, persis_info, gen_specs, libE_info=None):
-        self.persis_info = persis_info
+    def __init__(self, _, persis_info, gen_specs, libE_info=None, **kwargs):
         self.gen_specs = gen_specs
-        self.libE_info = libE_info
+        self.persis_info = persis_info
         self._get_user_params(self.gen_specs["user"])
+        self.gen_specs["out"] = [("x", float, (self.n,))]
```
shuds13, line 63: Not sure about this. We are trying to represent an external gen here. Most external gens are not going to have this.

libensemble/gen_classes/sampling.py
```diff
@@ -31,11 +31,10 @@ class UniformSample(SampleBase):
...
     """
 
-    def __init__(self, _, persis_info, gen_specs, libE_info=None):
-        self.persis_info = persis_info
-        self.gen_specs = gen_specs
-        self.libE_info = libE_info
+    def __init__(self, _=[], persis_info={}, gen_specs={}, libE_info=None, **kwargs):
+        super().__init__(_, persis_info, gen_specs, libE_info, **kwargs)
         self._get_user_params(self.gen_specs["user"])
+        self.gen_specs["out"] = [("x", float, (self.n,))]
```
shuds13, line 37: Interesting. This would overwrite any gen_specs['out'] supplied.

libensemble/generators.py
```diff
@@ -104,7 +117,7 @@ def ask(self, num_points: Optional[int] = 0) -> List[dict]:
 
     def tell(self, results: List[dict]) -> None:
         """Send the results of evaluations to the generator."""
-        self.tell_numpy(list_dicts_to_np(results))
+        self.tell_numpy(list_dicts_to_np(results))  # OH, we need the union of sim_specs.out and gen_specs.out
```
shuds13, line 120: If using externally may not even have sim_specs, or gen_specs.

libensemble/gen_classes/sampling.py
jlnav: Yeah, good point; I don't think `gen_specs["out"]` is necessary for this specific class

libensemble/gen_classes/sampling.py
jlnav: Yeah, declaring input/output types is often implied by libE's interface to be something you can do; and something we should wholeheartedly support. but for UniformSample, for example, does it create anything _other than_ `x`s ?

I was thinking that it seems it doesn't matter if you decide to uniformly sample thetas instead:

```python
gen_specs["out"] = ["theta", float, (3,)]
```

the fields are often _nonetheless_ hardcoded like:

```python
H_o["x"] = ...
```

Maybe that's just because its a very simple gen. But this is also true for our other gens; you really can't customize ``gen_specs["out"]`` except for sizes.

So might as well hardcode the gen's fields, I guess, and make sure its clear to a user that its really the size thats customizable.

Though I guess if fields really were configurable, we'd have a gen that resembles:

```python
for field in gen_specs["out"]:
   if field == "theta":
      H["theta"] = do_experiment_with_theta()
   else:
      H["x"] = do_without_theta()
```

libensemble/generators.py
jlnav: Exactly. I think we briefly discussed on Tuesday that we have to do the conversion anyway; and thankfully libE can handle receiving numpy arrays with dtypes it previously was unaware of

shuds13: For the external interface, I would think each gen must need to say name/type for dictionaries entries (in VOCS). So user knows what they must supply. But they only allow scalars and something like sampling can take varying dimensions, what would the interface be. x0, x1, x2 ... Xn

shuds13 (approved):

Issue #1404, opened 2024-08-14
Pydantic warnings


shuds13:
Tests in CI showing a lot of pydantic warnings such as:

/home/runner/miniconda3/envs/condaenv/lib/python3.9/site-packages/pydantic/_internal/_config.py:322: UserWarning: Valid config keys have changed in V2:
* 'orm_mode' has been renamed to 'from_attributes'
  warnings.warn(message, UserWarning)

See https://github.com/Libensemble/libensemble/actions/runs/10393072943/job/28779738691

jlnav:
Closed by #1410

PR #1410, opened 2024-08-20
bump pydantic versions, really trying to resolve warnings

jlnav:
Tries to address #1404

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -18,28 +18,28 @@ jobs:
                 os: [ubuntu-latest]
                 mpi-version: [mpich]
                 python-version: [3.9, "3.10", "3.11", "3.12"]
-                pydantic-version: ["2.6.4"]
+                pydantic-version: ["2.8.2"]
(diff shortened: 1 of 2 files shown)


jlnav:
orm_mode culprit: https://github.com/search?q=repo%3Aargonne-lcf%2Fbalsam%20orm_mode&type=code

root_validator culprit: https://github.com/search?q=repo%3Aargonne-lcf%2Fbalsam+%40root_validator&type=code

jlnav:
Seems to me that to test Balsam, to avoid the warnings we should only install it on the pydantic 1 jobs

shuds13 (commented):
This seems to only be changing what we test?

shuds13 (approved):
If the change is just only installing balsam with pydantic 1 testing, as its not compatible with pydantic 2, I approve.

jlnav:
Basically yeah, all I did was bump Pydantic versions then only install Balsam on Pydantic 1. It'll get tested on the Pydantic 1 jobs

PR #1419, opened 2024-08-29
Refactor/remove ytopt heffte

jlnav:
With ytopt/heffte moved to community examples, we can remove them in the main repo to cleanup / shorten builds, and not distribute as much code we don't use/support.

The `ytopt_asktell.py` gen is _not_ used upstream by the ytopt folks.


shuds13 (approved):
Approve if dont lose coverage

Issue #1403, opened 2024-08-14
Better capture dependencies for 'extra' tests.


shuds13:
The dependencies for CI tests should be captured in a more encapsulated way. This may means a `requirements.txt` file or `requirements_extra.txt` with all dependencies listed. 

This has to somehow be reconciled with the many conditions (on Python versions etc) that are currently in the yml file.

jlnav:
Closed by recent CI changes and the recent release

PR #1420, opened 2024-08-30
Removing dependency caching, some extra.yml dep installs into requirements.txts

jlnav:
Addresses #1403 

Plus some other refactors and cleanups.

Its difficult because we have combinations of OS, pydantic version, python version, etc.


jlnav:
I *believe* we knew this already, but installing gpcam uninstalls numpy 2

shuds13 (approved):

PR #1424, opened 2024-09-06
Refactor/pydantic adjusts

jlnav:
Pydantic v2 _and_ v1.10.17 have a `pydantic.v1` compatibility namespace/module that we can import everything we need from. This means that we only need a single interface (unfortunately, the old one) to be compatible with both versions.

This is much cleaner, but I wouldn't be surprised if there's a slight performance drawback if the validation in the `v1` module only occurs in python instead of rust.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -18,18 +18,18 @@ jobs:
                 os: [ubuntu-latest]
                 mpi-version: [mpich]
                 python-version: [3.9, "3.10", "3.11", "3.12"]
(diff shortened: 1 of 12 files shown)


jlnav:
TODO: fix docs

jlnav:
The docs issue may be related to autodoc_pydantic not supporting the `BaseModel` in `pydantic.v1` ? I opened an issue here to investigate: https://github.com/mansenfranzen/autodoc_pydantic/issues/302

PR #1429, opened 2024-09-13
First attempt at allowing decorating of objects with e.g. ``@output_data(["x"])``

jlnav:
So these fields show up in docs, but won't affect gen_specs


jlnav:
This is probably made redundant/unneeded by the standard requesting that gens explicitly sample over specified variables

PR #1437, opened 2024-10-07
Convert test runner to Python

shuds13:
Trying a Python version of the test runner.

Aim is to simplify and make it easier to see where in the script the tests are being run.
- [x] Test working with github Actions
- [x] Check works with codecov
- [x] Check works with extra tests
- [x] Check test vars are read correctly (e.g. TESTSUITE_EXTRA)
- [x] Compare runs  old and new scripts inc. which tests run/coverage/num procs 
- [x] Check special build code for forces.x
- [x] Check works right on macOS
- [x] Remove old bash test runner
- [x] Is clean function sufficiently safe
- [x] Does color work on both unit and regression tests.

Changes to tests / approach 
- [x] Do we really want to default to printing all test output or only on failure?*
- [x] Remove any commas in directives - e.g: # TESTSUITE_NPROCS: 5, 7
- [DEFERRED] Can .coveragerc files be consolidated and omit lists dont include all unit test dirs.
- [DEFERRED] Can conftest.py be consolidated (seems needs to be in current dir but might be alt config).
- [x] Remove some things that are not generally used like PEP option/python version/test strings/name.
- [x] Do we want to keep REG_RUN_LARGEST_TEST_ONLY which is currently the default.
- [x] Some tests were not being run as had no TESTSUITE directives. Setting default (e.g. local and 4 procs).
- [x] Should revisit test_persistent_uniform_gen_decides_stop.py due to intermittent failures.
- [x] test_1d_sampling_with_profile.py  takes a very long time for some reason. Reduce num sims/re-visit profiling?
- [DEFERRED] coverage looks like most of tasmanian gen not being tested.

*Example of run where I've turned the "show output" option off - shows stderr when test fails (could show stdout also):
https://github.com/Libensemble/libensemble/actions/runs/11229218671/job/31214403085


libensemble/tests/run_tests.py
```diff
@@ -0,0 +1,382 @@
...
+    """Run a shell command and display its output."""
+    if suppress_output:
+        with subprocess.Popen(
+            cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+        ) as proc:
+            stdout, stderr = proc.communicate()
+            if proc.returncode != 0:
+                raise subprocess.CalledProcessError(proc.returncode, cmd)
+    else:
+        print(f"\n{cmd=}\n")
```
shuds13, line 129: Remove

libensemble/tests/run_tests.py
```diff
@@ -0,0 +1,457 @@
...
+
+def run_command(cmd, cwd=None, suppress_output=False):
+    """Run a shell command and display its output."""
+    # print(f"\nCommand: {' '.join(cmd)}\n")  # For debugging
+    if suppress_output:
+        with subprocess.Popen(cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True) as proc:
+            stdout, stderr = proc.communicate()
+            if proc.returncode != 0:
+                # SH: Print error output if fail (could show stdout also?)
+                # print(stdout)
```
jlnav, line 165: Yeah, I think showing stdout on fail seems reasonable

jlnav (approved):

libensemble/tests/functionality_tests/test_persistent_uniform_gen_decides_stop.py
```diff
@@ -83,7 +83,7 @@
                 sum(counts == init_batch_size) >= ngens
             ), "The initial batch of each gen should be common among initial_batch_size number of points"
             assert (
-                len(np.unique(counts)) > 1
+                len(counts) > 1
```
shuds13, line 86: This assert, does not really match the comment. As it is the second "unique" it is really checking that successive generator cycles process diff num of points. To check for diff gen_ended_times - remove the unique (thats what I thought we were doing).

If you want to check after initial batch, you deal with different counts, I'm not sure you can easily ensure that with one gen. That is, if all sims come back, in a given manager cycle, it will be the same.

shuds13: We already check counts[0] == init_batch_size 
I can do
            for i in range(ngens):
                assert counts[i] == init_batch_size, (...
                
would that be sufficient ?

PR #1446, opened 2024-10-21
Refactor/deprecate gen dirs

jlnav:


jlnav:
Upon entering a generator's .ask, or .tell, possibly switch to gen directory

Issue #1444, opened 2024-10-16
Replace use of  pkg_resources


shuds13:
- Fix this: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html on install 

- May go with updating to use pyproject.toml

jlnav:
May be a good idea for v2.0; plus I believe lots of misc .dotfiles can be put in pyproject.toml too

jmlarson1:
Moving dotfiles content elsewhere could be nice.

jlnav:
I believe closed by recent adoption of pyproject.toml

PR #1459, opened 2024-11-05
Add pyproject.toml, with pixi as project management tool. add pixi.lock

jlnav:
Helps address #1444 and #1403

This adds a `pyproject.toml` to replace `setup.py`, which the Python community dramatically encourages.

For project management I chose Pixi (https://pixi.sh/latest/):

- Can install/interact with **both** conda and PyPI packages. For example, I can specify `mpich` as a dev dependency without having to say "install pydantic with pip and mpich with conda".
- Can split dependencies/packages into project-inherent "environments". Use `pixi install -e dev` to use the `dev` environment. With the lockfile included we're all _guaranteed_ to get a working environment since the lockfile was resolved earlier.
- Very fast rust backend.
- Installing a package documents/adds that install in the pyproject.toml.
- Platform-specific dependencies can be specified to only install on that platform. e.g. us linux people that install the project will get gpcam. As a macOS user I automatically get a different clang from conda-forge I use with mpicc.

This does not mean that users _need_ Pixi, and y'all don't _need_ it either.


jmlarson1:
I'd check the instances of `setup.py` being used/referred. From the root libE directory, run
````
grep -ir --include=*.* "setup\.py"
````

**Comment by jlnav** [comment]
Noting that a very minimal `setup.py` has been added back. But it'll never need to be updated.

**Comment by jlnav** [comment]
More information about `pixi.lock`:

https://pixi.sh/latest/features/lockfile/#how-to-use-a-lock-file

File: pyproject.toml [hunk 1]
```diff
@@ -0,0 +1,134 @@
+[project]
+
+authors = [{name = "Jeffrey Larson"}, {name = "Stephen Hudson"},
+            {name = "Stefan M. Wild"}, {name = "David Bindel"},
+            {name = "John-Luke Navarro"}]
+
+dependencies = ["pip>=24.3.1,<25", "setuptools>=75.1.0,<76", "numpy>=1.21,<3",
```
shuds13, line 7: "pip>=24.3.1,<25", "setuptools>=75.1.0,<76",
should prob be below in build systems so separate build from run deps.

pyproject.toml
```diff
@@ -0,0 +1,134 @@
...
+platforms = ["osx-arm64", "linux-64", "osx-64"]
+
+[tool.pixi.pypi-dependencies]
+libensemble = { path = ".", editable = true }
+
+[tool.pixi.environments]
+default = []
+dev = ["dev"]
+
+[tool.pixi.feature.dev.dependencies]
```
shuds13, line 60: Thinking lower bounds should be based on what we support. Or if we want to be latest, maybe make a comment to that effect.

shuds13:
I suggest we dont have lock file in here for now.

shuds13:
And as Jeff points out above:

```
(libe_general) shudson@shuds:libe_toml$ grep -ir --include=*.* "setup\.py"
libensemble/tests/unit_tests_logger/.coveragerc:    */setup.py/*
libensemble/tests/unit_tests_mpi_import/.coveragerc:    */setup.py/*
libensemble/tests/regression_tests/.coveragerc:    */setup.py/*
libensemble/tests/.coveragerc:    */setup.py
libensemble/tests/unit_tests/.coveragerc:    */setup.py/*
libensemble/tests/functionality_tests/.coveragerc:    */setup.py/*
libensemble/tests/unit_tests_nompi/.coveragerc:    */setup.py/*
MANIFEST.in:include setup.py
docs/xSDK_policy_compatibility.md:|**M1.** Support xSDK community GNU Autoconf or CMake options. |N/A| libEnsemble is a Python package and provides a `setup.py` file for installation. This is compatible with Python's built-in installation feature (`python setup.py install`) and with the ubiquitous `pip` installer. libEnsemble is also in the Spack repository and can be installed with `spack install py-libensemble`. GNU Autoconf or CMake are unsuitable for a Python package.
docs/xSDK_policy_compatibility.md:|**R5.** Provide a mechanism to export ordered list of library dependencies. |Full| The dependencies for libEnsemble are given in `setup.py` and when pip install or pip setup.py egg_info are run, a file is created `libensemble.egg-info/requires.txt` containing the list of required and optional dependencies. If installing through pip, these will automatically be installed if they do not exist (`pip install libensemble` installs req. dependencies, while `pip install libensemble[extras]` installs both required and optional dependencies.|
docs/dev_guide/release_management/release_platforms/rel_pypi.rst:In the package directory on the main branch (the one containing setup.py) do
docs/dev_guide/release_management/release_platforms/rel_pypi.rst:    python setup.py sdist
docs/dev_guide/release_management/release_process.rst:- ``MANIFEST.in`` is checked. Locally, try out ``python setup.py sdist``
```

shuds13 (approved):

PR #1465, opened 2024-11-13
Update gpCAM.py

roussel-ryan:
Fixes some small issues preventing tell in the generator standard

### DIFF
--- a/libensemble/gen_classes/gpCAM.py
+++ b/libensemble/gen_classes/gpCAM.py
@@ -74,7 +74,7 @@ class GP_CAM(LibensembleGenerator):
                 input_set=np.column_stack((self.lb, self.ub)),
                 n=n_trials,
                 pop_size=n_trials,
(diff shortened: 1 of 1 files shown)


libensemble/gen_classes/gpCAM.py
```diff
@@ -74,7 +74,7 @@ def ask_numpy(self, n_trials: int) -> npt.NDArray:
                 input_set=np.column_stack((self.lb, self.ub)),
                 n=n_trials,
                 pop_size=n_trials,
-                acquisition_function="total correlation",
```
shuds13, line 77: This should prob be passed as option, with sensible default.

shuds13: But also there is "MINIMIZE", "MAXIMIZE" keywords. If we added 'EXPLORE', that would probably be "total correlation".

Issue #1468, opened 2024-11-15
MPIExecutor wait_time always set by wait_on_start


cchall:
The change to the `wait_on_start` feature of the MPIExecutor in 1ba9f5721bcd7794b7a8c0ac6dfc81516c9f6c03 to change it from bool to either bool or int  looks like it results in MPIExecutor.fail_time always being ignored. 

In the case `wait_on_start=True` it looks like MPIExecutor.fail_time should be used to set the wait time. However, the check https://github.com/Libensemble/libensemble/blob/2892c864cc0a0d01e9ad6424d218b44ef2433640/libensemble/executors/mpi_executor.py#L169
always passes with the first condition because bool is a subclass of int. This can cause failed simulations to not be caught if fail_time > 1 was needed.

jlnav:
Thanks Chris for the bug report! The corresponding fix should appear in our upcoming release.

PR #1226, opened 2024-01-25
Fixes/from 112 optimas

jlnav:
- `wait_on_start=N`: N seconds of waiting before continuing
- Upon exception, `task.state = "FAILED"`

PR #1474, opened 2024-11-21
Fix wait_on_start type-instance condition checking to use fail_time w…

jlnav:
Addresses #1468.

Basically `True` and `False` got interpreted as `0` and `1`, valid integers for `wait_on_start`,
meaning the `Executor.fail_time` attribute never got used.

Now we use that attribute if we _actually_ got `True`. Any other integer >1 means we wait for that time. `False` being `0` should simply mean we don't wait.

Thanks @cchall!

### DIFF
--- a/libensemble/executors/executor.py
+++ b/libensemble/executors/executor.py
@@ -687,7 +687,7 @@ class Executor:
         stdout: Optional[str] = None,
         stderr: Optional[str] = None,
         dry_run: Optional[bool] = False,
-        wait_on_start: Optional[Union[bool, int]] = False,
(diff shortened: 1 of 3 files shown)


shuds13:
This used to be a boolean, what is the use case for waiting N seconds?

jlnav:
I don't exactly remember; I introduced the change in https://github.com/Libensemble/libensemble/pull/1226.

Based on grepping through _all of github_ and known frameworks that use libE, I can probably change the option back to boolean only.

shuds13 (approved):

PR #1486, opened 2024-12-06
Logging updates

shuds13:
Showing worker node with local comms can confuse users.

Only show worker node in the log when running in a distributed mode (e.g., MPI comms).

Also only report manager/worker initialization with debug logging.

This also adds a `VDEBUG` logging level for logging of communicating log messages, as this is rarely needed and is very verbose.

Removes unhelpful Uneven groups warning

### DIFF
--- a/libensemble/logger.py
+++ b/libensemble/logger.py
@@ -8,9 +8,11 @@ MANAGER_WARNING = 35
 logging.addLevelName(MANAGER_WARNING, "MANAGER_WARNING")
 logging.MANAGER_WARNING = MANAGER_WARNING
 
+
 def manager_warning(self, message: str, *args, **kwargs) -> None:
     if self.isEnabledFor(MANAGER_WARNING):
         self._log(MANAGER_WARNING, message, args, **kwargs)
+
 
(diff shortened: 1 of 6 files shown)


jlnav (approved):

libensemble/logger.py
```diff
@@ -15,6 +15,19 @@ def manager_warning(self, message: str, *args, **kwargs) -> None:
...
 logging.Logger.manager_warning = manager_warning
+
+VDEBUG = 5
+logging.addLevelName(VDEBUG, "VDEBUG")
+logging.VDEBUG = VDEBUG
+
+
+def vdebug(self, message, *args, **kwargs):
+    if self.isEnabledFor(VDEBUG):
+        self._log(VDEBUG, message, args, **kwargs)
```
jmlarson1, line 26: CodeCov says that this line isn't covered, but I would think that the `test_logger.py` unit test would cover it.

shuds13: Test added.

shuds13: Test really added.

jmlarson1 (approved):

PR #1489, opened 2024-12-13
Reduce dependabot commits via grouping and lowering frequency

shuds13:
Dependabot is swamping developer commits.


jlnav:
Very good idea

jmlarson1 (approved):

PR #1501, opened 2025-01-07
Docs/generator section refactor

jlnav:


docs/examples/sampling.rst
```diff
@@ -29,5 +29,5 @@ persistent_sampling
 persistent_sampling_var_resources
 ---------------------------------
 .. automodule:: persistent_sampling_var_resources
-  :members:
+  :members: uniform_sample_with_var_gpus
```
shuds13, line 32: We should include all the gen_f functions unless they are very obviously just for testing and/or would be misleading to users. It is good to have a long list of functions, not bad. There are some good examples of using variable resources here in this file.

shuds13:
Some alternatives.

**Have a first page under `Examples:`** which has some curated information (more similar to what's in the drop-down box) and does not have to list every gen_f. Then “Generator Functions” only has a short description under sections.

**Try inserting short descriptions for each generator file or gen_f.**

**Use a table** instead  which gives greater flexibility to how we can express various information about the generators. This could be one table or it could be a separate table for different sections (in which case the sections would still appear in the left-hand banner)

docs/examples/aposmm.rst
```diff
@@ -12,32 +12,26 @@ Optional (see below): petsc4py_, nlopt_, DFO-LS_
...
 
-where ``optimizers`` is a string (or list of strings) from the available options:
+where ``optimizers`` is a string (or list-of-strings) from:
 
 ``"petsc"``, ``"nlopt"``, ``"dfols"``, ``"scipy"``, ``"external"``
 
-.. dropdown:: Issues with ensemble hanging or failed simulations?
+.. dropdown:: Issues with ensemble hanging or failed simulations with PETSc?
 
-    Note that if using **mpi4py** comms, PETSc must be imported at the global
```
shuds13, line 29: removed?

docs/examples/aposmm.rst
```diff
@@ -12,32 +12,26 @@ Optional (see below): petsc4py_, nlopt_, DFO-LS_
 Configuring APOSMM
 ^^^^^^^^^^^^^^^^^^
 
-APOSMM works with a choice of optimizers, some requiring external packages. To
-import the optimization packages (and their dependencies) at a global level
-(recommended), add the following lines in the calling script before importing
-APOSMM::
+APOSMM works with a choice of optimizers, some requiring external packages. Specify
```
shuds13, line 15: there are occasions where we cant do this

jlnav:
Maybe now that the README may have:

```
libEnsemble empowers model-driven ensembles on the world's leading supercomputers
including Frontier, Aurora, and Perlmutter.
```

We can drop this line for simplicity:

```
libEnsemble is effective at solving design, decision, and inference problems on parallel resources.
```

jlnav:
Maybe update this line on github?:

```A Python toolkit for coordinating asynchronous and dynamic ensembles of calculations. ```

jlnav:
- Probably remove Overview of Examples page (or comment-out)
- Simpler Ax multipoint gen or example
- Alphabetize within categories
- Sampling, (basic sampling)

jlnav:
- Have drop-down in quickstart default to open

jlnav:
Perhaps move "see the tutorial" and "see user guide" under Support or Further Information

wildsm:
In quickstart, maybe combine the sentences around the bullets into one:
_libEnsemble empowers model-driven ensembles to solve design, decision, and inference problems on the world’s leading supercomputers including Frontier, Aurora, and Perlmutter._

jlnav:
If possible, little stubs to discuss each sim_f

jlnav:
"forces" sims are templates rather than "forces" sims?

docs/examples/aposmm.rst
shuds13: https://github.com/Libensemble/libensemble/pull/1064

jmlarson1: Should we add a note to point to this issue with PETSc?

jmlarson1 (approved):

PR #1515, opened 2025-02-18
Return None from persistent gens

shuds13:
- [x] What about APOSMM and fd_param_finder?

### Explanation

Look at this code, which is typical for our persistent gens.

```python

    while tag not in [STOP_TAG, PERSIS_STOP]:
        H_o = np.zeros(b, dtype=gen_specs["out"])
        H_o["x"] = persis_info["rand_stream"].uniform(lb, ub, (b, n))
        tag, Work, calc_in = ps.send_recv(H_o)


    return H_o, persis_info, FINISHED_PERSISTENT_GEN_TAG

```

H_o is sent to the manager in `ps.send_recv(H_o)`. Then at end those points are returned (again). They get added twice - the manager will give new sim_ids. We decided to ignore the return H_o unless `use_persis_return_gen` is set, in order to prevent that. Also, bear in mind, users will have used these as templates to their own gens. 

We should make our gens end with 

`return None, persis_info, FINISHED_PERSISTENT_GEN_TAG`

unless we do something after the loop and actually make a different H_o.

The same should be done in community examples.

We can do this with the intent to drop  `use_persis_return_gen` once this is normalized.

### DIFF
--- a/libensemble/gen_funcs/persistent_ax_multitask.py
+++ b/libensemble/gen_funcs/persistent_ax_multitask.py
@@ -305,7 +305,7 @@ def persistent_gp_mt_ax_gen_f(H, persis_info, gen_specs, libE_info):
         # Increase iteration counter.
         model_iteration += 1
 
-    return [], persis_info, FINISHED_PERSISTENT_GEN_TAG
+    return None, persis_info, FINISHED_PERSISTENT_GEN_TAG
 
 
 class AxRunner(Runner):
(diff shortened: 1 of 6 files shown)


shuds13:
> I think this is great for gens that are communicating "everything" back to the manager along the way.
> 
> For other gens that have quantities of their own that aren't being communicated back during the run, I think we still want to be able to return H_o

I have only returned None in cases where the returned values would be a simple duplicate of what's already been returned.

PR #1547, opened 2025-04-09
Testing/some coverage + another attempt at dynamic versioning

jlnav:


libensemble/ensemble.py
```diff
@@ -578,14 +578,11 @@ def save_output(self, basename: str, append_attrs: bool = True):
         Format: ``<basename>_results_History_length=<length>_evals=<Completed evals>_ranks=<nworkers>``
         """
         if self.is_manager:
-            if self._get_option("libE_specs", "workflow_dir_path"):
```
shuds13, line 581: I think this change should wait as its a code change post testing

libensemble/resources/platforms.py
```diff
@@ -289,12 +289,10 @@ def known_envs():
             logger.manager_warning("Perlmutter detected, but no compute partition detected. Are you on login nodes?")
     if os.environ.get("SLURM_CLUSTER_NAME") == "lumi":
         partition = os.environ.get("SLURM_JOB_PARTITION")
-        print(f"Lumi partition: {partition}")
         if not partition:
             logger.manager_warning("LUMI detected, but no compute partition detected. Are you on login nodes?")
         if partition and partition.endswith("-g"):
             name = "lumi_g"
-            print(f"Lumi GPU detected: {name}")
```
shuds13, line 297: good catch

libensemble/gen_funcs/persistent_tasmanian.py
```diff
@@ -10,10 +10,10 @@
 from libensemble.tools import parse_args
 from libensemble.tools.persistent_support import PersistentSupport
 
-__all__ = [
```
shuds13, line 13: I dont think we should remove this- doesn't the docs read this

shuds13 (approved):

libensemble/gen_funcs/persistent_tasmanian.py
jlnav: I had to verify this when I was puzzling why that line wasn't covered before, but it turns out `__all__` only gets covered when one does:

`from some_module import *`

This isn't done in our code, or the docs:

```
.. automodule:: persistent_tasmanian
  :members: sparse_grid_batched, sparse_grid_async
  :undoc-members:
```

PR #1548, opened 2025-04-10
Replace flashing live animation with clean post-run version

shuds13:
https://colab.research.google.com/github/Libensemble/libensemble/blob/examples/aposmm_nb_animation/examples/tutorials/aposmm/aposmm_tutorial_notebook.ipynb

2D or 3D:

3D plot better shows where minima are, but takes longer to render..
Also markers dont show up as well (hence the Ms)


jlnav (approved):
Very cool!

PR #1552, opened 2025-04-22
Refactor/310 plus typehints

jlnav:


shuds13 (commented):
Lets try out mypy with some tests

jlnav:
A bunch of the above commits are from mypy's feedback. mypy is _really_ strict about making sure types, variables, outputs, etc. are exactly as expected, and all discrepancies are accounted for. For instance, class-attributes that default to None, that are always not-None from our experience, mypy points out that we need to more carefully assert that those attributes aren't None before we try accessing them.

libensemble/gen_funcs/persistent_ax_multitask.py
```diff
@@ -45,19 +43,36 @@
     AxParameterWarning = Warning
 
 from ax.modelbridge.factory import get_sobol
-from ax.modelbridge.registry import MBM_X_trans, Models, ST_MTGP_trans
+from ax.modelbridge.registry import Models, ST_MTGP_trans
```
shuds13, line 46: check why this is changing

PR #1556, opened 2025-04-30
Testing/pixi in testing

jlnav:


shuds13 (approved):

jlnav:
At this time, foot in mouth moment, the time spent to get *only* gpcam compatible with the pixi testing environment may be substantially longer then it'll be worth converting the environment at this time. I may come back to this in the future

PR #1562, opened 2025-05-23
Example showing a search over rk

jmlarson1:


jmlarson1:
Need to adjust/add to the APOSMM doc string to point to this example.

Issue #1555, opened 2025-04-30
Drop Pydantic v1 support


jlnav:

jlnav:
Closed by #1573

PR #1573, opened 2025-07-03
Refactor/remove pydantic 1 and balsam

jlnav:
#1555


jmlarson1:
Is there any reason for having the "if sphinx" [part here](https://app.codecov.io/gh/Libensemble/libensemble/pull/1573/blob/libensemble/utils/pydantic_bindings.py) in line 62?

jlnav:
> Is there any reason for having the "if sphinx" [part here](https://app.codecov.io/gh/Libensemble/libensemble/pull/1573/blob/libensemble/utils/pydantic_bindings.py) in line 62?

The `create_model` function updates our specs with the correct validators, but doesn't play nice when autodoc-pydantic tries to render those specs; the fields don't get listed. This is the recommended solution.

PR #1588, opened 2025-09-05
Swap default alloc_f to only_persistent_gens

jlnav:
- [ ] Perhaps deprecate simple, non-persistent gens
- [ ] Adjust tutorials
- [ ] Adjust docs descriptions of gens
- [ ] Perhaps, for many tests that just use give_sim_work_first for sampling, refactor for persistent_sampling


jlnav:
Perhaps this should wait until ask/tell gets pulled in.

PR #1622, opened 2025-11-20
Remove automapping

shuds13:
Experimenting with removing automapping.

xopt test runs with this change.

We should run all tests and see what breaks. 

Do we need to keep auto-mapping, in an optional sense, or modify it.

If remove will have to update unit tests.

Ive also added `gest_api_wrapper` sim func so xopt/gest-api style sims can be used directly


shuds13:
✔️ test_xopt_EI.py
✔️ test_asktell_aposmm_nlopt.py
✔️ test_asktell_gpCAM.py
❌ test_asktell_sampling.py (relies on automapping)

PR #1640, opened 2025-12-18
start_only_persistent alloc_f is now the default alloc

jlnav:
Used to target develop. Now targets asktell-develop, with adjustments to the included regression tests.


libensemble/alloc_funcs/give_sim_work_first.py
```diff
@@ -53,8 +52,11 @@ def give_sim_work_first(
     if libE_info["sim_max_given"] or not libE_info["any_idle_workers"]:
         return {}, persis_info
 
-    # Initialize alloc_specs["user"] as user.
-    batch_give = user.get("give_all_with_same_priority", False)
+    # Initialize options - check gen_specs first
```
shuds13: The alloc specs is created as user but not used anywhere here. also docstrings still says it uses alloc_specs but it only uses gen_specs. I think the way I would do it would be keep with user and just make it both. E.g.,

`user = {**gen_specs.get("user", {}), **alloc_specs.get("user", {})}`

then all other lines can be unchanged from original, right?

this can be done for all allocs, unless they have alloc/gen specs that conflict.

shuds13:
are names consistent across allocs?
do we want to reconsier async_return name?

libensemble/specs.py
```diff
@@ -233,6 +233,44 @@ class GenSpecs(BaseModel):
...
+    """
+    Create gen in active receive mode. If True, the manager does not need to wait
+    for a return from the generator before sending further returned points.
+    Default: False. Only used if using the ``only_persistent_gens`` allocation function (the default).
+    """
+
+    give_all_with_same_priority: bool = False
+    """
+    If True, then all points with the same priority value are given as a batch to the sim.
+    Default: False. Only used if using the ``only_persistent_gens`` allocation function (the default).
```
shuds13, line 259: is it not also in give_sim_work_first.

shuds13 (approved):

PR #1650, opened 2026-01-13
Fix App Check: Container Path

ax3l:
When using the `precedent` to run apps from container images, the executable is not available to check for the host side. This relaxes respective "early" checks to support this workflow.

X-ref: https://github.com/optimas-org/optimas/pull/292


ax3l:
@shuds13 the remaining test failure looks unrelated to this PR.

shuds13 (approved):

ax3l:
@shuds13 oh, I fear I need this PR as well. Can we finalize it together?

Issue #1633, opened 2025-12-08
Incrementally adding mypy to the libEnsemble Project


the-mr-dave:
In discussion with the maintainer @jlnav the lack of static type checking in the project accured. We propose that mypy will be set up for the project and the type checking will be incrementally be extended to the whole project.

First step would be a setup, that won't disrupt the current CI.

jlnav:
Closed by recent PR

PR #1655, opened 2026-01-14
Add initial mypy configuration and CI integration

the-mr-dave:
# Description
This PR adds mypy type checking to libEnsemble using an intentionally permissive initial configuration.
It establishes a baseline that can be tightened incrementally without impacting CI stability.

Related issue: #1633

# Changes
- Add an initial mypy configuration to pyproject.toml
- Add mypy as a CI check
- Restrict type checking scope to libensemble/utils
- Exclude utility modules that are not yet mypy-compliant
- Refactor libensemble/utils/misc.py to satisfy mypy

# CI Impact
- mypy is executed with a restricted scope and permissive settings
- No existing CI checks are modified, relaxed, or made less strict

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -63,6 +63,12 @@ jobs:
             pip install -e .
             flake8 libensemble
 
+        - name: Install mypy
+          run: pip install mypy
+
+        - name: Run mypy (limited scope)
+          run: mypy
+
         - name: Remove various tests on newer pythons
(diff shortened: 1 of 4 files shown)


jlnav:
Hello,

Thank you for the mypy implementation and fixes. The configuration and CI implementation look good.

However, I'm surprised changes to `output_directory.py` were included even though it's excluded by mypy's checks. Did you find a bug? If so, how did you come across it?

shuds13 (approved):

PR #1676, opened 2026-02-12
Set executor base dir to ensemble dir

shuds13:


ax3l (approved):
Awesome! This successfully removes the static `exploration_dir_path` from the `LIBENSEMBLE_SIM_DIR` for me 🚀

PR #1690, opened 2026-03-06
Docs/cleanup of overview

jmlarson1:


jlnav (approved):
Looks good to me, but as we work on 2.0 we'll need to look into referring to gens as "generators" instead of "gen_f"s as much as possible

shuds13 (approved):

Issue #1525, opened 2025-02-27
only_persistent_gens becomes default alloc for 2.0


jlnav:

jlnav:
Closed by #1702

PR #1702, opened 2026-03-26
Refactor/new alloc default tweaks

jlnav:
- Remove globus_compute_endpoint from gen_specs. Registering a function and running remotely makes no more sense with how gens have changed.
- Rename "give_all_with_same_priority" -> "batch_evaluate_same_priority"
- Place "generator" higher than "gen_f" in GenSpecs class. It will get listed first.
- Some docstring tweaking and rewording.

Follow-on to PR that addresses #1525


libensemble/specs.py
```diff
@@ -243,42 +241,39 @@ class GenSpecs(BaseModel):
...
-    If True, then the specialized allocator behavior for some persistent gens is used.
-    Only used if using the ``only_persistent_gens`` allocation function (the default).
+    Enable specialized allocator behavior for ``only_persistent_gens``.
     """
 
     batch_mode: bool = False
     """
-    If True, then the generator will not be started if there are still simulations
-    running. Only used if using the ``give_sim_work_first`` allocation function.
+    Don't query the generator until all running simulations have finished.
```
shuds13: Lets only duplicate to gen_specs the default alloc options, so this does not appear as an option. And if listed under alloc options should keep the Only used if using the ``give_sim_work_first`` allocation function.

shuds13 (approved):

PR #1711, opened 2026-04-11
VOCS APOSMM direct implementation

shuds13:


jlnav:
It was always a matter of time :)

shuds13:
We should have the ability to not have to pass *on_cube around.

PR #1715, opened 2026-04-14
remove use_persis_return_sim plus mypy fixes

jlnav:


shuds13:
Is the persistent sim example still getting tested. 

Remove the sim and gen versions of this variable - check no persistent sim or gen is returning data that is not needed (ie. dupcliate of whatever was returned in loop). When they should return, you can see something with update H after loop.

shuds13 (approved):

PR #1728, opened 2026-05-14
Add vocs sampling gens

shuds13:
- [x] Adds remaining sampling generators as gest-api gens.
- [x] Updates `initial_sample_method` to accept sample constructed with options
- [x] Keeps string option for `initial_sample_method` for simple samplers.
- [x] Adds test for LHS class and using `initial_sample_method` with sampler instance.

Remaining methods to be tested when tests are bulk updated to use VOCS.


jlnav:
I'm assuming when tests are bulk-updated is when the old versions will be deprecated/deleted?

jlnav (approved):

PR #1731, opened 2026-05-19
Feature/updated globus compute

jlnav:
Older worker-local Globus-Compute-submissions removed, replaced with:

GlobusComputeExecutor

- Function-submitter that largely mirrors the interface of the other libEnsemble executors

GlobusCompute "Mode"

- The generator is local to the manager by default, so _other workers aren't technically needed to run sims if using globus compute_.
- Sim work, with all corresponding inputs, is submitted directly by the manager to globus compute.
- "legacy" and "new"-styled sims supported.
- "nworkers" controls number of simultaneous parallel submissions to globus compute.
- Output/input work is wrapped as though it went-to/came-from a worker

Credit to Claude, of course


jlnav:
Closing due to having split this into 3 branches / PRs

PR #1737, opened 2026-05-25
Add libE service mode

shuds13:
Allow libE to act as a dispatch engine for an external producer.

Typical examples may include MCP servers, LLM agents, REST services, message-queue consumers, or an interactive environment.

### DIFF
--- a/docs/data_structures/libE_specs/libE_specs_general.rst
+++ b/docs/data_structures/libE_specs/libE_specs_general.rst
@@ -38,3 +38,17 @@ General
 **gen_workers** [list of ints]:
     List of workers that should run only generators. All other workers will run
     only simulator functions.
+
+**service_mode** [bool] = ``False``:
+    If ``True``, the manager tolerates idle workers + an alloc returning no work
+    (instead of asserting). Used when libEnsemble is driven by an external
+    producer (e.g. a queue-backed Generator fed by an MCP server) and may
+    legitimately have nothing to dispatch for periods of time. The manager
+    sleeps briefly between checks instead of panicking. Termination is the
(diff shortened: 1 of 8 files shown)


docs/data_structures/libE_specs/libE_specs_general.rst
```diff
@@ -38,3 +38,17 @@ General
 **gen_workers** [list of ints]:
     List of workers that should run only generators. All other workers will run
     only simulator functions.
+
+**service_mode** [bool] = ``False``:
+    If ``True``, the manager tolerates idle workers + an alloc returning no work
```
jlnav, line 43: I think maybe this docstring could emphasize usage and implications more than internal implementation.

e.g. "If ``True`` libEnsemble waits for work idly, like from queue-backed generators or MCP servers, instead of exiting immediately if work isn't available. Termination is still the user's responsibility, via "exit_criteria" or other external signals."

jlnav:
I can also imagine pushing "sim work" onto a queue for libE to evaluate in parallel. But I guess if the gen grabs that works and gives to libE, the same thing is accomplished.

libensemble/tests/regression_tests/test_queue_gen.py
```diff
@@ -0,0 +1,66 @@
...
+# TESTSUITE_COMMS: local
+# TESTSUITE_NPROCS: 4
+
+import math
+import time
+
+import numpy as np
+from gest_api.vocs import VOCS
+
+from libensemble.gen_classes.queue_gen import QueueService
```
jlnav, line 21: Perhaps a different namespace like `libensemble.services.QueueService` ?

PR #1759, opened 2026-07-08
Fix mypy type errors in mpi_resources.py

jlnav:


libensemble/resources/mpi_resources.py
```diff
@@ -252,12 +260,13 @@ def create_machinefile(
...
             logger.warning(f"Could not remove existing machinefile: {e}")
 
+    assert resources.worker_resources is not None  # for mypy
     node_list = resources.worker_resources.local_nodelist
     logger.debug(f"Creating machinefile with {num_nodes} nodes and {procs_per_node} ranks per node")
 
     with open(machinefile, "w") as f:
         for node in node_list[:num_nodes]:
-            f.write((node + "\n") * procs_per_node)
+            f.write((node + "\n") * (procs_per_node or 1))
```
shuds13, line 269: check this is right. As procs_per_node is not always set by user. Is it filled in at this point. Is this default correct.

libensemble/resources/mpi_resources.py
```diff
@@ -102,29 +102,31 @@ def task_partition(
     """
 
     # Convert to int if string is provided
-    num_procs = int(num_procs) if num_procs else None
-    num_nodes = int(num_nodes) if num_nodes else None
-    procs_per_node = int(procs_per_node) if procs_per_node else None
+    num_procs_int: int | None = int(num_procs) if num_procs else None
```
shuds13, line 105: why _int?

PR #1784, opened 2026-09-16
[1/4] Add test runner controls and timing reports

jlnav:
- Adds coverage toggles, fail-fast, test filters, list mode, sharding, and slow-test timing reports.
- Validates standalone `TESTSUITE_*` metadata and adds focused runner unit tests.
- Replaces the duplicated deduplication proposal with a measured test-acceleration strategy.
- Stack: parent #1780; child #1783.


jlnav:
Follow-up fix pushed in commit `096c5761d`: added `mypy >=1.19.1,<2` to the shared Pixi `basic` feature and regenerated `pixi.lock`. Verified `pixi run -e py312 mypy` succeeds with no issues. The combined local Flake8 command still reports an unrelated warning in the pre-existing untracked `libensemble/tests/unit_tests/test_features.py` file (missing final newline); that file was left untouched.

